PR #9, opened 2020-06-23
Create home page.

Sami-2000:
Create a home page file, currently just including a basic header.

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>Open Sesame</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="script.js"></script>
+  </head>
+  <body>
+    <h1>The Open Source Community is for <span class="emphasis">Everyone</span></h1>
+  </body>
+</html>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -7,6 +7,6 @@
     <script src="script.js"></script>
   </head>
   <body>
-    <h1>The Open Source Community is for <span class="emphasis">Everyone</span></h1>
+    <h1>The Open Source Community is for <span id="emphasis">Everyone</span></h1>
   </body>
 </html>
--- a/src/main/webapp/style.css
+++ b/src/main/webapp/style.css
@@ -1,0 +1,7 @@
+h1 {
+  font-size: x-large;
+}
+
+.emphasis {
+  color: violet;
+}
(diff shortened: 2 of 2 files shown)


src/main/webapp/index.html
```diff
@@ -0,0 +1,12 @@
...
+  <head>
+    <meta charset="UTF-8">
+    <title>Open Sesame</title>
+    <link rel="stylesheet" href="style.css">
+    <script src="script.js"></script>
+  </head>
+  <body>
+    <h1>The Open Source Community is for <span class="emphasis">Everyone</span></h1>
+  </body>
+</html>
```
Richie78321, line 12: nit: Add newline at end of file

src/main/webapp/style.css
```diff
@@ -0,0 +1,7 @@
+h1 {
+  font-size: x-large;
+}
+
+.emphasis {
+  color: violet;
+}
```
Richie78321, line 7: nit: Add newline at end of file

Obinnabii:
```suggestion
}

```

src/main/webapp/index.html
Obinnabii:
```suggestion
</html>

```

src/main/webapp/style.css
Richie78321, line 7: Please look into [Bootstrap](https://getbootstrap.com/) to increase consistency among all of our website. For example, consider using [the display class](https://getbootstrap.com/docs/4.0/content/typography/#display-headings) from Bootstrap instead of defining a larger font size for the header.

src/main/webapp/style.css
```diff
@@ -0,0 +1,7 @@
+h1 {
+  font-size: x-large;
+}
+
+.emphasis {
+  color: violet;
```
Richie78321, line 6: Also, if you think that we'll be moving forward with violet as our accent color, it might be worth looking into [theming Boostrap](https://getbootstrap.com/docs/4.0/getting-started/theming/#modify-map) to change their built-in color scheme to match what we're looking for.

Richie78321: Although I've never worked with Sass before so maybe leave this as a nice-to-have for now. Don't need to overcomplicate things.

Obinnabii: yeah I think making a class with CSS should be fine.

Obinnabii: It get's the job done and you can always call it if the need arises

Richie78321: I agree. Just throwing out options b/c Bootstrap is cool

Sami-2000: @Richie78321 what is Sass? Also I will def look into Bootstrap for the next iteration if that will make our site more consistent. If you want to go with general theming (which looks really cool) maybe we can decide on the specifics during Pod Sunk?

src/main/webapp/style.css
Sami-2000: Will do!

PR #10, opened 2020-06-24
Ignore JDK files

Obinnabii:
update the gitignore file to ignore JDK files as well. JDK is useful for local Java programming on chromebooks. The meta data created by JDK might cause errors when working in non-Linux environments. The gitignore additions are taken from [here](https://github.com/Obinnabii/my-portfolio/blob/master/.gitignore)  

@billyeh linked me to a page with a fully set up gitignore [here](https://github.com/redhat-developer/vscode-java/blob/master/.gitignore). The link above is to the gitignore that worked on my portfolio.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,42 +6,6 @@
 # Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
 # Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
 
-# User-specific stuff
-.idea/**/workspace.xml
-.idea/**/tasks.xml
-.idea/**/usage.statistics.xml
-.idea/**/dictionaries
-.idea/**/shelf
-
-# Generated files
-.idea/**/contentModel.xml
-
-# Sensitive or high-churn files
-.idea/**/dataSources/
-.idea/**/dataSources.ids
-.idea/**/dataSources.local.xml
-.idea/**/sqlDataSources.xml
-.idea/**/dynamic.xml
-.idea/**/uiDesigner.xml
-.idea/**/dbnavigator.xml
-
-# Gradle
-.idea/**/gradle.xml
-.idea/**/libraries
-
-# Gradle and Maven with auto-import
-# When using Gradle or Maven with auto-import, you should exclude module files,
-# since they will be recreated, and may cause churn.  Uncomment if using
-# auto-import.
-# .idea/artifacts
-# .idea/compiler.xml
-# .idea/jarRepositories.xml
-# .idea/modules.xml
-# .idea/*.iml
-# .idea/modules
-# *.iml
-# *.ipr
-
 # CMake
 cmake-build-*/
 
--- a/.gitignore
+++ b/.gitignore
@@ -125,7 +125,6 @@ hs_err_pid*
 
 ### JDK ###
 .idea
-out
 server
 *.vsix
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -127,12 +127,9 @@ hs_err_pid*
 .idea
 out
 server
-node_modules
 *.vsix
 .DS_Store
-.vscode-test
 undefined
-dist
 bin/
 .settings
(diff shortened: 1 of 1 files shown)


Obinnabii:
@googlebot I signed it!

.gitignore
```diff
@@ -123,6 +123,22 @@ fabric.properties
...
 hs_err_pid*
 
+### JDK ###
+.idea
+out
+server
+node_modules
+*.vsix
+.DS_Store
+.vscode-test
```
Obinnabii, line 133:
```suggestion
```

.gitignore
```diff
@@ -123,6 +123,22 @@ fabric.properties
...
+undefined
+dist
```
Obinnabii, line 135:
```suggestion
```

.gitignore
```diff
@@ -123,6 +123,22 @@ fabric.properties
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+### JDK ###
+.idea
+out
+server
+node_modules
```
Obinnabii, line 130:
```suggestion
```

.gitignore
Obinnabii: duplicate

.gitignore
Obinnabii: duplicate

.gitignore
Obinnabii: duplicate

.gitignore
```diff
@@ -123,6 +123,19 @@ fabric.properties
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+### JDK ###
+.idea
+out
```
Richie78321, line 128:
```suggestion
```
This already exists in the gitignore

.gitignore
```diff
@@ -123,6 +123,19 @@ fabric.properties
 # virtual machine crash logs, see http://www.java.com/en/download/help/error_hotspot.xml
 hs_err_pid*
 
+### JDK ###
+.idea
```
Richie78321, line 127:
```suggestion
```
This is handled above as well (IntelliJ)

.gitignore
```diff
@@ -123,6 +123,19 @@ fabric.properties
...
+out
+server
```
Richie78321, line 129: Looking at gitignore patterns [here](https://www.atlassian.com/git/tutorials/saving-changes/gitignore), writing just `server` means that it matches both files named `server` and directories named `server`. Is this intended?

.gitignore
```diff
@@ -123,6 +123,19 @@ fabric.properties
...
+*.vsix
+.DS_Store
+undefined
```
Richie78321, line 132: See above

.gitignore
Obinnabii: yeah

.gitignore
Obinnabii: the idea tags above are intellij specific

Obinnabii: and won't get rid of certain files

Richie78321:
Oh okay. I looked at the sources again and it seems like you're getting it directly from vscode-java, so I won't bother with the specifics. Sorry for going to hard on that lol.

.gitignore
Obinnabii: This gitignore file worked for me previously. If it has any unintended effects we can revise it later but I don't think there should be any issues.I'd like to avoid having random computer generated files on git. That is just prone to causing errors. For example, in the portfolio project, I ran into an problem where my code was compiling with errors because of an old target directory. Not sending these files to other peoples systems will probably be the best course of action. Also, for context, this is an official JDK gitignore file.

Obinnabii:
@Richie78321 does you're account have CLA approval? Google bot is being weird.

Richie78321:
@googlebot I fixed it.

.gitignore
billyeh: If you want to just ignore .idea/, I suggest removing the above redundant lines that are more specific e.g. 5-41, 49, 64, 92, etc.

PR #12, opened 2020-06-24
Add basic React navbar component

Richie78321:
Adds a basic React navbar component that can be easily added to any page by including the React scripts in the HTML, as well as the `Navbar.js` script.

Also revises the `package.json` file to add the common React components directory `js/common/react` to the JSX preprocessor. Along with this, a shortcut `npm run server-dev` is added to make starting the dev server from cloud shell easier by combining the JSX preprocessing step.


Richie78321:
Going to look into style for JSX and reopen

Richie78321:
Going to wrap lines according to the [HTML style guide](https://google.github.io/styleguide/htmlcssguide.html#HTML_Line-Wrapping) to stay within the JS 80 column limit.

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Open Sesame is a platform intended to make contributing to open source less intimidating by matching first-time contributors to experienced mentors.",
   "main": "index.js",
   "scripts": {
-    "jsx-preprocessor": "npx babel --watch src/main/webapp/js/projects/react/ --out-dir src/main/webapp/js/projects/react/jsx-processed --presets react-app/prod"
+    "dev-server": "npm run jsx-preprocessor && mvn package appengine:run || true",
```
Sami-2000, line 7: Just a Q: What is the benefit of including the || true?

src/main/webapp/js/common/react/Navbar.js
```diff
@@ -0,0 +1,60 @@
...
+  
+  return (
+    <div className={classes}>
+      <a className="nav-link" href={props.href}>{props.name}</a>
+    </div>
+  );
+}
+
+const navbarContainer = document.getElementById("navbar-container");
+ReactDOM.render(<Navbar />, navbarContainer);
```
Sami-2000, line 60: Just a Q: wondering how this rendering gets called? (doesn't immediately appear to be part of a function)

package.json
Richie78321: Good eye! It's actually pretty interesting. `||` essentially means that if the command to the left throws an error (starting the dev server) then run the command to the right, which is in this case `true` and pretty much just evaluates to nothing. This is used because otherwise every time you did Ctrl+C to close the dev server npm would throw an error.

src/main/webapp/js/common/react/Navbar.js
Richie78321: This is called when the script is loaded as a part of the global scope. And if you're wondering about motivation, it allows the navbar to be loaded when the script is included, and avoids having to call any loading functions. I'd like to hear Bill or Joe's opinion on if this is considered smelly code tho.

package.json
billyeh: Not very obvious so it's worth commenting. JSON doesn't support comments, maybe we can throw a little Bash comment in the string?
```suggestion
    "dev-server": "npm run jsx-preprocessor && mvn package appengine:run || true  # Don't throw an error after exiting",
```

billyeh:
I can see this coming in handy!

README.md
```diff
@@ -2,15 +2,27 @@
...
+```
+npm install
+```
+
+**Before running a development server or deploying to production:**
+```
+npm run jsx-preprocessor
+```
+
+* **There is a nifty shortcut** included that runs the JSX preprocessor and then immediately starts the development server from the cloud shell. Use this by running `npm run dev-server` in the outermost directory.
```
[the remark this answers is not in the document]
Richie78321: Yup. Good catch!

PR #14, opened 2020-06-25
Add button to homepage.

Sami-2000:
Use bootstrap for the title. Add a button with a shadow on hover (will link to form, currently links nowhere).

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -4,6 +4,6 @@
 <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
 <link rel="stylesheet" href="style.css">
 <script src="script.js"></script>
-<h1 class="display-1">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
+<h1 class="display-1"><div id="home_header">The Open Source Community is for <span class="emphasis">Everyone</span></div></h1>
 <spacer class="space"></spacer>
-<button class="form_button">Mentorship Awaits</button>
+<button class="btn">Mentorship Awaits</button>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -10,6 +10,6 @@
   <body>
     <h1 class="display-1">The Open Source Community is for <span id="emphasis">Everyone</span></h1>
     <spacer class="space"></spacer>
-    <button class="form_button">Mentorship Awaits</button>
+    <button id="form_button">Mentorship Awaits</button>
   </body>
 </html>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -6,4 +6,4 @@
 <script src="script.js"></script>
 <h1 class="display-1">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
 <spacer class="space"></spacer>
-<button id="form_button">Mentorship Awaits</button>
+<button class="form_button">Mentorship Awaits</button>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -4,6 +4,7 @@
 <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
 <link rel="stylesheet" href="style.css">
 <script src="script.js"></script>
+
 <h1 class="display-1 mb-5 mt-5">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
 <spacer class="space"></spacer>
 <button class="btn mt-5 text-center"><span class="display-4">Mentorship Awaits</span></button>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -1,15 +1,9 @@
 <!DOCTYPE html>
-<html>
-  <head>
-    <meta charset="UTF-8">
-    <title>Open Sesame</title>
-    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
-    <link rel="stylesheet" href="style.css">
-    <script src="script.js"></script>
-  </head>
-  <body>
-    <h1 class="display-1">The Open Source Community is for <span id="emphasis">Everyone</span></h1>
-    <spacer class="space"></spacer>
-    <button id="form_button">Mentorship Awaits</button>
-  </body>
-</html>
+<meta charset="UTF-8">
+<title>Open Sesame</title>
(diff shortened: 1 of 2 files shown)


src/main/webapp/style.css
```diff
@@ -1,8 +1,22 @@
...
+
 h1 {
-  font-family: Verdana, sans-serif; 
-  font-size: x-large;
+  margin-top: 100px; 
 }
 
-#emphasis {
-  color: violet;
+.button {
```
Obinnabii, line 9: It might be good to give this a more specific name or to style the type button if this will be the default button look.

src/main/webapp/style.css
```diff
@@ -1,8 +1,22 @@
...
+  border: none;
+  border-radius: 8px;
+  font-size: xx-large;
+  left: 50%;
+  margin-top: 200px;
+  text-align: center;
+  width: 70%;
 }
+
+.button:hover {
```
Obinnabii, line 20: See comment  on line 9

Obinnabii:
```suggestion
button:hover {
```

src/main/webapp/index.html
```diff
@@ -3,10 +3,13 @@
...
     <link rel="stylesheet" href="style.css">
     <script src="script.js"></script>
   </head>
   <body>
-    <h1>The Open Source Community is for <span id="emphasis">Everyone</span></h1>
+    <h1 class="display-1">The Open Source Community is for <span id="emphasis">Everyone</span></h1>
+    <spacer class="space"></spacer>
+    <button class="button">Mentorship Awaits</button>
   </body>
 </html>
```
Obinnabii, line 15: some tags don't do anything such as ```<html>, <head> and <body>``` the style guide talks a bit about removing these tags [here](https://google.github.io/styleguide/htmlcssguide.html#Optional_Tags). the documentation isn't super clear so it is an optional style requirement. I think code is a bit easier to read without the ```<html>``` and ```<head>``` tags. This is completely optional.

src/main/webapp/style.css
Sami-2000: Sounds good! I will make it a more specific id.

src/main/webapp/style.css
```diff
@@ -1,8 +1,22 @@
+#emphasis {
+  color: violet;
+}
+
 h1 {
```
Richie78321, line 5: Looking at [the CSS style guide](https://google.github.io/styleguide/htmlcssguide.html#Type_Selectors), it seems like we should be avoiding using type selectors like `h1` for CSS. I also think it's important to remember that this will apply to every `h1` element that uses this stylesheet, which I don't think is what we want. Please consider using [Bootstrap margins](https://getbootstrap.com/docs/4.0/utilities/spacing/)

src/main/webapp/style.css
```diff
@@ -1,8 +1,22 @@
...
+#form_button {
+  background-color: violet;
+  border: none;
+  border-radius: 8px;
+  font-size: xx-large;
+  left: 50%;
+  margin-top: 200px;
+  text-align: center;
+  width: 70%;
 }
```
Richie78321, line 18: Please look to use Bootstrap where at all possible with styling. Consider using `mt-4` or similar for margins. And `text-center` for centering text. Also consider maybe using the existing [Bootstrap buttons](https://getbootstrap.com/docs/4.0/components/buttons/) and overriding the color differences for consistency. I plan to use Bootstrap buttons so we should aim to keep it consistent.

src/main/webapp/style.css
Sami-2000: You asked me to use the display-1 from bootstrap for the heading. Because display-1 is pre-defined in bootstrap, I cannot change it to add a top margin. How do you recommend I make a heading which is h1, display-1, and also has a custom margin, without using type selectors?

src/main/webapp/style.css
Sami-2000: The Bootstrap buttons are very limited. I wanted to stay consistent with our color scheme, and think the button is much more appealing with the curved corners and shadow on hover. That is why I did not choose to use a Bootstrap button.

src/main/webapp/index.html
Obinnabii: This was the link Bill showed me for the window.onload functionality. https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onload

src/main/webapp/style.css
Sami-2000: Also, is it not nicer to have a single, well-defined form_button id rather than stacking a bunch of Bootstrap classes (a basic bootstrap button, mt-4, text-center) and my own modifications?

src/main/webapp/index.html
Sami-2000: Thank you! I def feel comfortable removing the optional tags now :)

src/main/webapp/style.css
Obinnabii: Maybe we could try using an ID for this particular header?

src/main/webapp/style.css
Obinnabii: hmmm. I think it would be worth it to have a default button theme. I think Richie is trying to avoid losing out on some of the bootstrap functionality. If we use the ```btn``` class and add our own class with the modifications that it doesn't have.
for instance;
```
button.btn {
  border-radius: 8px;
  text-align: center;
}
```
You could add the ```!important``` tag to override bootstrap if the changes don't show too.

src/main/webapp/style.css
Sami-2000: Good idea @Obi! Trying that now.

src/main/webapp/style.css
Sami-2000: Thank you so much @Obi! That sounds like an awesome plan.

Sami-2000: I am going to push a new version. Let me know if this is what you intended @Obinnabii

src/main/webapp/style.css
Richie78321: You can apply multiple Bootstrap classes to the same element, so I imagine something like this would work:
```
<h1 class="display-1 mt-4">[...]</h1>
```

src/main/webapp/style.css
```diff
@@ -1,8 +1,22 @@
...
+  font-size: xx-large;
+  left: 50%;
+  margin-top: 200px;
+  text-align: center;
+  width: 70%;
+}
+
+button.btn:hover {
+  box-shadow: 0 12px 16px 0 rgba(0,0,0,0.24),0 17px 50px 0 rgba(0,0,0,0.19);
+}
```
Richie78321, line 22: +1 for box shadows. They don't get enough love.

src/main/webapp/index.html
```diff
@@ -4,6 +4,6 @@
 <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
 <link rel="stylesheet" href="style.css">
 <script src="script.js"></script>
-<h1 class="display-1">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
+<h1 class="display-1"><div id="home_header">The Open Source Community is for <span class="emphasis">Everyone</span></div></h1>
```
Obinnabii, line 7: maybe adding the ID to the h1 tag.
```suggestion
<h1 id= "home_header" class="display-1">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
```

Sami-2000: Wow I had no clue you could have an id and a class on the same <>!!! Thanks

src/main/webapp/style.css
Obinnabii: yeah it looks good! does it work when you preview?

src/main/webapp/index.html
Obinnabii: no problem!

Obinnabii:
LGTM!

Richie78321:
Yeah this looks great. Thanks for your patience with us Sami.

PR #15, opened 2020-06-25
Dashboard Prototype Framework 

Obinnabii:
- Create a dashboard page
- Add Bootstrap 4 tab functionality. I plan to use tabs to simplify how the mentors view mentees and projects.
- Add some minor styling to the tab headers.


src/main/webapp/dashboard.html
```diff
@@ -0,0 +1,53 @@
...
+    </li>
+    <li class="nav-item">
+      <a class="nav-link" data-toggle="tab" href="#int-mentees">Interested Mentees</a>
+    </li>
+  </ul>
+
+  <!-- Tab panes -->
+  <div class="tab-content">
+    <!-- About Me -->
+    <div id="abt-me" class="container tab-pane active"><br>
```
Richie78321, line 32: Should probably avoid abbreviations where possible to preserve readability. Could change to `about-me`.

src/main/webapp/dashboard.html
```diff
@@ -0,0 +1,53 @@
...
+      <h3>About Me</h3>
+      <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
+    </div>
+    <!-- Affiliated Projects -->
+    <div id="affiliations" class="container tab-pane fade"><br>
```
Richie78321, line 37: This ID could be more specific. How about something like `project-affiliations`?

src/main/webapp/dashboard.html
```diff
@@ -0,0 +1,53 @@
...
+      <h3>Affiliations</h3>
+      <p>Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
+    </div>
+    <!-- Interested Mentees -->
+    <div id="int-mentees" class="container tab-pane fade"><br>
```
Richie78321, line 42: Same comment as above. Should avoid the abbreviation with `interested-mentees`.

src/main/webapp/style.css
```diff
@@ -1,8 +1,14 @@
 h1 {
-  font-family: Verdana, sans-serif; 
+  font-family: Verdana, sans-serif;
   font-size: x-large;
 }
```
Richie78321, line 4: Consider creating a class for this styling to [avoid using type selectors](https://google.github.io/styleguide/htmlcssguide.html#Type_Selectors) or consider applying the font to a more global element like `body` to keep the font consistent between different text elements.

src/main/webapp/style.css
```diff
@@ -1,8 +1,14 @@
...
-#emphasis {
+.emphasis,
+a {
   color: violet;
 }
+
+a.active,
+a:hover {
+  color: rebeccapurple !important;
+}
```
Richie78321, line 14: nit: no newline

src/main/webapp/style.css
```diff
@@ -1,8 +1,14 @@
...
 
-#emphasis {
+.emphasis,
+a {
   color: violet;
 }
+
+a.active,
+a:hover {
+  color: rebeccapurple !important;
```
billyeh, line 13: Just wondering why we need the !important?

src/main/webapp/style.css
Obinnabii: Yeah def, I think this is here from rebasing on another branch. I’ll fix it!

src/main/webapp/dashboard.html
Obinnabii: I was planning on showing mentors/mentees that you have matched with her too

src/main/webapp/style.css
Obinnabii: The color wasn't showing up without it for some reason. I think the bootstrap styling was overriding it.

src/main/webapp/dashboard.html
Richie78321: Oh okay. Then I think maybe the comment above that div is a bit misleading.

src/main/webapp/style.css
Richie78321: Should have been more specific -- no newline at the end of the file.

Obinnabii:
```suggestion
}

```

PR #18, opened 2020-06-25
Add a basic project search frontend with mock data

Richie78321:
Adds a basic project search frontend with mock data.

* Add `ProjectPreview` component, which is a card-based component that displays only the most relevant information of a project to be used in the search UI.
* Add `ProjectList` component, which creates a list of `ProjectPreview` components from inputted project preview data
  * This component will be expanded in the future as the `ProjectPage` component, which will handle the loading of project data and search input alongside listing the project previews.
* Add `ProjectTagList` component, which creates a visualization of project tags.
  * This is intended to be used in both the preview and whole-page versions of a project's UI, which is why it is contained in its own component.

*Also adds a live JSX Preprocessor for the projects react components to support a live preview server*


billyeh:
Could you please merge master branch and we'll get some linting on the JS files? :) I'm a little scared, it might be a bit aggressive.

Richie78321:
> Could you please merge master branch and we'll get some linting on the JS files? :) I'm a little scared, it might be a bit aggressive.

Do you mean merge this branch with master or merge something to master?

billyeh:
> > Could you please merge master branch and we'll get some linting on the JS files? :) I'm a little scared, it might be a bit aggressive.
> 
> Do you mean merge this branch with master or merge something to master?

Merge master into this branch.

Richie78321:
@billyeh Did it! It seems like it doesn't support JSX 100% (it doesn't see that I'm using imported components that are being used in JSX) but it isn't too bad.

Richie78321:
@billyeh It seems like getting rid of [no-undef] and [no-unused-imports] would solve most of the problems. Thoughts? If that seems reasonable I'll ignore them in the linter.

src/main/webapp/js/common/react/Navbar.js
```diff
@@ -1,36 +1,36 @@
...
-      <button 
-          className="navbar-toggler" 
-          type="button" 
-          data-toggle="collapse" 
-          data-target="#navbarSupportedContent" 
-          aria-controls="navbarSupportedContent" 
-          aria-expanded="false" 
+      <button
+          className="navbar-toggler"
+          type="button"
```
Sami-2000, line 23: Is the type attribute necessary? Not sure exactly how the style sheets apply to react, but they say to "omit type attributes." https://google.github.io/styleguide/htmlcssguide.html#HTML_Style_Rules

src/main/webapp/projects.html
```diff
@@ -3,13 +3,28 @@
   <head>
```
Sami-2000, line 3: Consider removing optional tags, as suggested by the html style sheet: https://google.github.io/styleguide/htmlcssguide.html#HTML_Style_Rules

src/main/webapp/css/projects.css
```diff
@@ -0,0 +1,9 @@
+.project-preview {
+  transition: .05s;
+}
+
+.project-preview:hover {
+  box-shadow: 0px 0px 5px 1px rgba(232,232,232,1);
+  -moz-box-shadow: 2px 2px 5px 1px rgba(232,232,232,1);
```
Sami-2000, line 7: Should we consider removing the box-shadow prefixes?: https://css-tricks.com/do-we-need-box-shadow-prefixes/
(seems like less than 0.3% of Firefox and Safari users are still on versions that require the prefix)

Sami-2000: Also super super nit but the ordering of declarations as described on the style sheet seems to show that vendor-specific prefixed versions of a property should come before the non-prefixed property https://google.github.io/styleguide/htmlcssguide.html

src/main/webapp/js/common/react/Navbar.js
Richie78321: Yeah good catch. I took that Navbar boilerplate directly from Bootstrap so it seems like it is missing some of Google's style. Will update!

src/main/webapp/css/projects.css
Richie78321: Yeah I think you're right about that. So I think then the rule is vendor-specific come before but alphabetized. Nice!

Richie78321: And that first article you sent is great. I agree it seems like they're unnecessary. Will remove.

src/main/webapp/projects.html
Richie78321: I'll definitely take a look at this, as I see both you and Obi are doing it and it'd probably be best if we stayed consistent. It hurts my heart a bit as it looks quite different from what I'm used to, but I'll look to change it in the future.

PR #20, opened 2020-06-26
Add React support to ESLint

Richie78321:
Adds React support to ESLint with this npm package: https://www.npmjs.com/package/eslint-config-react-app.

Adds the browser environment to the linter to allow for use of `window` and `document`.

Adds `React` and `ReactDOM` as globals to avoid triggering the [no-undef] error.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,9 +2,20 @@
   "extends": [
     "eslint:recommended",
     "google",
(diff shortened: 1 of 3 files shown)


Richie78321:
To see it in action on a more up to date branch, take a look at #18

PR #22, opened 2020-06-26
Create frontend of mentor search page to support simplified mentors.

Sami-2000:
Add simplified (only name and description) mentors to cards on a basic mentor search page.


Obinnabii:
fixed a merge conflict

src/main/webapp/mentors.js
```diff
@@ -0,0 +1,47 @@
+/**
+ * Get mentors from the mentor servlet.
+ */
+function getMentors() {
+  console.log("entering get mentors function/n");
+  let url = new URL('/mentors', location.protocol + '//' + location.hostname);
+
+  fetch(url).then(response => response.json()).then(mentors => {
```
Richie78321, line 8: Look to add error handling to this fetch request in the future. If you add a `.catch((err) => [...])` after the fetch it should cover when there is an error making the request (like a network error). Take a look at [promise catches](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/catch) for more context. And then from within the response (before converting to json), you can add a check to determine if the request was successful (as determined by the response's [HTTP code](https://www.restapitutorial.com/httpstatuscodes.html)). This is a nice shortcut:
```
.then((response) => {
  if (response.ok) {
  [...]
  }
}
```
Where the response is okay if it is a 2xx code. The one thing that is a bit messy is what to do when the response is not okay. There are a bunch of ways you could handle this. If you handle the error from within that if-statement and don't return anything, it is the same as returning a resolved promise with a value of undefined, so it will get to the next then block with the json, but in this case the json will be undefined so you could then check if there was an error with something like:
```
if (json) {
[...]
}
```

Thanks for listening to my TED talk!

src/main/webapp/mentors.js
```diff
@@ -0,0 +1,47 @@
...
+      mentorsContainer.appendChild(createMentorElement(mentor.name, mentor.description));
+    }
+  });
+}
+
+/**
+ * Creates a mentor container element from input text.
+ * @param {String} name
+ * @param {String} description
+ * @return {<div class="p-1 col-lg-4">} mentorContainer
```
Richie78321, line 22: I'm not sure if this is valid JSDoc. Consider replacing with `Element`. For context take a look at the return type for `document.createElement()` [here](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElement).

src/main/webapp/style.css
```diff
@@ -17,5 +17,12 @@ button.btn {
...
   box-shadow: 0 12px 16px 0 rgba(0,0,0,0.24),0 17px 50px 0 rgba(0,0,0,0.19);
+}
+
+.card-holder {
+  transition: .05s;
+}
 
+.card-holder:hover {
+  box-shadow: 0 12px 16px 0 rgba(0,0,0,0.24),0 17px 50px 0 rgba(0,0,0,0.19);
 }
```
Richie78321, line 28: Thanks for making a generalized version! Will add to mine in my next PR.

src/main/webapp/mentors.js
Obinnabii: Yeah, I think it should be ```HTMLElement```

src/main/webapp/style.css
Obinnabii: Yeah! I didn't notice that! Will use!

src/main/webapp/mentors.js
Sami-2000: Thanks for the advice and concrete examples!

src/main/webapp/mentors.js
```diff
@@ -0,0 +1,47 @@
...
+  const mentorName = document.createElement('h5');
+  mentorName.className = 'card-title text-primary';
+  mentorName.innerHTML = name;
+  mentorCardBody.appendChild(mentorName);
+
+  const mentorDescription = document.createElement('p');
+  mentorDescription.innerHTML = description;
+  mentorCardBody.appendChild(mentorDescription);
+
+  mentorCard.appendChild(mentorCardBody);
```
Obinnabii, line 43: It is usually better to define things when you use them. I think it might be better to create ```mentorCard``` and ```mentorContainer``` at the end, right before they are used or to do all the appending for these at the top before appending the child ```div```s

src/main/webapp/mentors.js
```diff
@@ -0,0 +1,47 @@
...
+  mentorContainer.appendChild(mentorCard);
```
Obinnabii, line 44: nit: It is usually better to define things when you use them. I think it might be better to create ```mentorCard``` and ```mentorContainer``` at the end, right before they are used or to do all the appending for these at the top before appending the child ```div```s

PR #23, opened 2020-06-26
Mentors backend 1

Sami-2000:
Create basic mentor class and mentor servlet to send a list of hard-coded, simplified (only name and description) mentors.

### DIFF
--- a/src/main/java/MentorObject.java
+++ b/src/main/java/MentorObject.java
@@ -1,0 +1,16 @@
+package com.google.sps.servlets;
+
+public class MentorObject {
+  private String name;
+  private String description;
+  public MentorObject(String nameInput, String descriptionInput) {
+    name = nameInput;
+    description = descriptionInput;
+  }
+  public String getName() {
+    return name;
+  }
+  public String getDescription() {
+    return description;
+  }
+}
--- a/src/main/java/MentorsServlet.java
+++ b/src/main/java/MentorsServlet.java
@@ -1,0 +1,27 @@
+package com.google.sps.servlets;
+
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.net.MalformedURLException;
+import java.net.URL;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.util.ArrayList;
+
+@WebServlet("/mentors")
+public class MentorsServlet extends HttpServlet {
+
+  @Override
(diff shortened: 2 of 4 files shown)


Richie78321:
nit: Would you be able to update the directories of these servlet files so they comply with [Java's directory convention for packages](https://docs.oracle.com/javase/tutorial/java/package/managingfiles.html)? For reference, this is also the structure used in our portfolios. Should be `webapp/java/com/google/opensesame/`.

src/main/java/MentorObject.java
```diff
@@ -0,0 +1,19 @@
+package com.google.sps.servlets;
+
+public class MentorObject {
+  private String name;
+  private String description;
```
Richie78321, line 5: We should discuss how we want to structure our data before we commit to this, but this definitely works for now. Take a look at what we have in the data doc we made yesterday and feel free to add your thoughts.

Richie78321: For example, I'm wondering if the `name` property should be part of a more general User class, as both mentors and mentees will have names.

src/main/java/MentorsServlet.java
```diff
@@ -0,0 +1,24 @@
+package com.google.sps.servlets;
```
Richie78321, line 1: Same as above comment
```suggestion
package com.google.opensesame.servlets;
```

src/main/java/MentorsServlet.java
```diff
@@ -0,0 +1,24 @@
...
+
+@WebServlet("/mentors")
+public class MentorsServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    ArrayList<MentorObject> mentors = new ArrayList<MentorObject>();
+    mentors.add(new MentorObject("Obi", "Obi is awesome."));
+    mentors.add(new MentorObject("Sami", "Sami is fun."));
+    mentors.add(new MentorObject("Richi", "Richi is cool."));
```
Richie78321, line 19: I see we're committing to Richi lol

Sami-2000:
Thanks for pointing out the com/google structuring! Are you sure the java folder goes in the webapp folder though? (totally cool if u r, but I recall that java was directly in main for the portfolio project)

src/main/java/MentorObject.java
Sami-2000: For sure! Let's touch base during sunk.

Richie78321:
> Thanks for pointing out the com/google structuring! Are you sure the java folder goes in the webapp folder though? (totally cool if u r, but I recall that java was directly in main for the portfolio project)

Oh yeah my bad you're right about that. Crap sorry.

PR #26, opened 2020-06-27
Add React prop type validation

Richie78321:
Adds [prop type validation](https://reactjs.org/docs/typechecking-with-proptypes.html) to existing React components.

### DIFF
--- a/src/main/webapp/js/common/react/Navbar.js
+++ b/src/main/webapp/js/common/react/Navbar.js
@@ -1,9 +1,8 @@
 /**
  * Returns a React navbar component.
- * @param {{}} props The properties of the navbar.
  * @return {React.Component} Returns the navbar.
  */
-function Navbar(props) {
+function Navbar() {
   const urls = [
     {
       href: '/',
@@ -54,6 +53,11 @@ function NavbarLink(props) {
     </div>
   );
 }
+
+NavbarLink.propTypes = {
+  href: PropTypes.string.isRequired,
+  name: PropTypes.string.isRequired,
+};
 
 const navbarContainer = document.getElementById('navbar-container');
 ReactDOM.render(<Navbar />, navbarContainer);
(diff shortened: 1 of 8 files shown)


src/main/webapp/js/common/react/Navbar.js
```diff
@@ -55,5 +54,10 @@ function NavbarLink(props) {
   );
 }
 
+NavbarLink.propTypes = {
```
Obinnabii, line 57: Could we define this in prop_types? or maybe define the other prop_types where they are used. I'm not completely sure what is happening but, to me it looks like you are defining types to avoid errors with user input. If I am right, I think that makes a lot sense. having a defined way to do this that doesn't vary with the file might be beneficial.

Obinnabii:
Besides the point above. LGTM!

src/main/webapp/js/common/react/Navbar.js
Richie78321: Indirectly this is somewhat validating user input, however the main feature of Prop Types is validating the [props](https://www.w3schools.com/react/react_props.asp) of a component to make sure that everything needed for the component to function is being inputted and everything is of the correct type. 

As for defining this in `prop_types.js`, I see what you're saying, but I think it'd be best if we followed the React conventions [which you can see on the React page on Prop Types](https://reactjs.org/docs/typechecking-with-proptypes.html). This includes setting the `propTypes` from within the file that defines the component.

Obinnabii: That makes sense! I saw that the project preview prop types were done differently, is there a reason for this?

Richie78321: Ahh good catch. So to better explain what I'm thinking, I'd like to keep `propType` assignments within each component class. The reason that the Project components are different is because there is a lot of repeat there with prop types as data is sent down from parent to child components. So to follow DRY, I added the repeating prop types to a separate file that each imports individually.

PR #27, opened 2020-06-28
Add React testing to Navbar components

Richie78321:
Adds support for testing React with [Jest](https://jestjs.io/). The tests can be run with `npm test`. Tests for the Navbar component were added to start.

For simplicity, these tests make use of:
https://testing-library.com/docs/react-testing-library/intro
https://github.com/testing-library/jest-dom

One thing that I would like feedback on is the introduction of conditional `require` statements, which you can see in `checkTesting.js`. This function needs to be called from any React component that is tested. This isn't ideal, but it seems
that the way we are using React is rather unorthodox ([adding React to a website](https://reactjs.org/docs/add-react-to-a-website.html)), and after a lot of digging this seems to be the best way to get testing to work.
For more detail, `checkTesting.js` is checking if the components are being run in a NodeJS environment, which would only be happening if they were being tested. Because the way we use React relies on globally-scoped React libraries imported from `<script>` tags in the HTML, these must be supplemented for while testing by importing them through Node.


src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -0,0 +1,37 @@
...
+  });
+
+  it('contains both urls', () => {
+    render(<Navbar urls={navbarUrls} />);
+
+    expect(screen.getByText('Home')).toHaveAttribute('href', '/');
+    expect(screen.getByText('Projects'))
+        .toHaveAttribute('href', '/projects.html');
+  });
+});
```
Obinnabii, line 37: This is such a cool testing environment

Richie78321: Ikr! Jest is great.. very readable.

src/main/webapp/projects.html
```diff
@@ -27,7 +27,7 @@
     <!-- Note: when deploying, replace "prop-types.js" with "prop-types.min.js" -->
     <script src="https://unpkg.com/prop-types@15.6/prop-types.js"></script>
     <!-- Load React components -->
-    <script type="module" src="./js/common/react/jsx-processed/Navbar.js"></script>
+    <script type="module" src="./js/common/react/jsx-processed/display_navbar.js"></script>
```
Obinnabii, line 30: Is there documentation for importing the navbar? It might be good to have a comment in the file explaining what link to add.

Richie78321: That would make sense. Will add some basic instructions in the `display_navbar.js` file.

src/main/webapp/js/common/react/tests/checkTesting.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Checks if the code is currently being run in a NodeJS environment, in which
+ * case the code is being tested and the React modules should be imported.
```
billyeh, line 3: Let's document some of our offline conversation here. I.e. why this is needed: these imports don't work in the browser. And possible solutions: bundling (Webpack, Browserify).

billyeh: I think you added a little description in the PR, but still didn't explain fully why we need these imports (just a simple "imports don't work in the browser" might have been enough). Both the comments in this file and the PR do a great job explaining *what* the code is doing. Though I can also figure that out by reading 4 lines of code :)

src/main/webapp/js/common/react/display_navbar.js
```diff
@@ -0,0 +1,30 @@
+/**
+ * @fileoverview Adds a React-based Navbar to an HTML page.
+ * Before importing this script, be sure to add the React scripts
+ * as seen here: https://reactjs.org/docs/add-react-to-a-website.html#step-2-add-the-script-tags.
+ * Also, be sure to add a div with the ID "navbar-container". This
+ * is where the Navbar will be rendered to.
+ */
+
```
billyeh, line 8: Let's address the lint warning here.

src/main/webapp/js/common/react/tests/checkTesting.js
Richie78321: Good point. The _what_ versus the _why_. I added a bit more context for the why and some of the alternatives considered.

PR #28, opened 2020-06-29
Add Js Functions for getting Dashboard Data

Obinnabii:
- My formatter was using four spaces. I changed the settings.
- Add function to add the ```About me card``` to the dashboard.

### DIFF
--- a/src/main/webapp/dashboard.js
+++ b/src/main/webapp/dashboard.js
@@ -1,7 +1,7 @@
 /**
  * Div of the About me tab.
  */
-const aboutMeDiv = document.getElementById('about-me')
+const aboutMeDiv = document.getElementById('about-me');
 
 /**
  * Fake information for the hardcoded stage.
@@ -13,7 +13,7 @@ const dummyBio = 'The cutest pug in The tech industry. Lorem ipsum dolor sit ' +
   ' labore et dolore magna aliqua.';
 const dummyImg = 'images/dior.jpg';
 const dummyTags = ['Weekly', '30 mins', 'Kubernetes', 'Open Sesame',
-  'Documentation'
+  'Documentation',
 ];
 const dummyGitSrc = '#';
 
@@ -35,7 +35,8 @@ function createAboutMe(name, location, bio, imgSrc, tags, gitSrc) {
   AboutMeCard.append(AboutMeCardBody);
 
   let userImage = document.createElement('img');
-  userImage.className = 'card-img-top img-thumbnail w-25 p-2 m-2 mx-auto d-block';
+  userImage.className = 'card-img-top img-thumbnail w-25 p-2 m-2 mx-auto' +
+    ' d-block';
   userImage.src = imgSrc;
   AboutMeCardBody.append(userImage);
 
--- a/src/main/webapp/dashboard.js
+++ b/src/main/webapp/dashboard.js
@@ -45,29 +45,25 @@ const dummyUser = {
 };
 
 /**
- * Populates the card element 'ABOUT_ME_CARD_DIV' with information about a
+ * Populate the card element 'ABOUT_ME_CARD_DIV' with information about a
  * given user.
- * @param {String} name
- * @param {String} location
- * @param {String} bio
- * @param {String} imgSrc
- * @param {String List} tags
+ * @param {User} user
  */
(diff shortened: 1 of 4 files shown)


src/main/webapp/dashboard.js
```diff
@@ -0,0 +1,94 @@
...
+  let AboutMeCard = document.createElement('div');
+  AboutMeCard.className = 'card card-holder col-12';
+
+  let AboutMeCardBody = document.createElement('div');
+  AboutMeCardBody.className = 'card-body';
+  AboutMeCard.append(AboutMeCardBody);
+
+  let editButton = document.createElement('a');
+  editButton.className = 'float-right bold';
+  editButton.innerText = 'Edit'
```
Sami-2000, line 39: Missing a ; (wow did the linter catch this too!?)

src/main/webapp/dashboard.js
```diff
@@ -0,0 +1,94 @@
...
+  editButton.href = '#'
```
Sami-2000, line 40: Another semicolon

src/main/webapp/dashboard.js
Sami-2000: Also, just a Q, but do you always want an about me card to always exist. jut with customized content? (if so, you could hard code the about me card in the page html, and only use a function like this to change the contents instead of creating a new elements. May make the function a bit shorter?)

Sami-2000: I'm just thinking about which of these divs could already exist, versus which you really need to create in this java script function.

src/main/webapp/style.css
```diff
@@ -1,23 +1,23 @@
...
-    color: darkblue;
+  color: darkblue;
 }
 
 .btn {
-    background-color: rebeccapurple;
-    border: none;
-    border-radius: 8px;
-    width: 70%;
+  background-color: rebeccapurple;
```
Sami-2000, line 13: Love Rebecca purple <3

src/main/webapp/dashboard.js
Obinnabii: yeah that makes a lot of sense. I'll re-evaluate my divs. I'll probably create helper functions as well to improve readability

Obinnabii:
```suggestion
  editButton.innerText = 'Edit';
```

src/main/webapp/dashboard.js
Obinnabii:
```suggestion
  editButton.href = '#';
```

Obinnabii:
@Sami-2000, I made some changes. Is this what you had in mind?

src/main/webapp/dashboard.js
```diff
@@ -0,0 +1,99 @@
...
+  let editButton = createEditButton('#');
+  ABOUT_ME_CARD_DIV.prepend(editButton);
+
+  USER_IMAGE.src = profileImgSrc;
+
+  USER_NAME_AND_LOCATION.innerHTML = name + '<br>';
+  let userLocation = createLocation(location);
+  USER_NAME_AND_LOCATION.append(userLocation);
+
+  USER_BIO.innerText = bio
```
Obinnabii, line 46:
```suggestion
  USER_BIO.innerText = bio;
```

src/main/webapp/dashboard.js
```diff
@@ -0,0 +1,99 @@
...
+const dummyName = 'Dior';
+const dummyLocation = 'Sunnyvale CA.';
+const dummyBio = 'The cutest pug in The tech industry. Lorem ipsum dolor sit ' +
+  'amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut' +
+  ' labore et dolore magna aliqua.';
+const dummyImg = 'images/dior.jpg';
+const dummyTags = ['Weekly', '30 mins', 'Kubernetes', 'Open Sesame',
+  'Documentation',
+];
+const dummyGitSrc = '#';
```
josephdhurst, line 25: You may want to store this data in the same format that you expect to get from the github API (I assume it would be a list of json objects but you should double-check). That way you'll be able to write the logic to iterate over a list of users before trying to integrate the github API with this code.

Obinnabii: will do!

PR #32, opened 2020-06-29
Add some robustness to mentor class; create person class.

Sami-2000:
For now I only added the functions that I immediately needed, and only the most basic parameters. Would appreciate any advice on handling optional inputs in java (so that I don't need a million constructors).


src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -1,19 +1,43 @@
...
-public class MentorObject {
-  private String name;
-  private String description;
+import java.util.ArrayList;
 
-  public MentorObject(String nameInput, String descriptionInput) {
-    name = nameInput;
-    description = descriptionInput;
+public class MentorObject extends PersonObject {
+  private ArrayList<String> projectIDs;
```
Richie78321, line 6: Nit: I feel that this name could be more descriptive. Consider something like `mentoringProjectIds`.

src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -1,19 +1,43 @@
...
+
+  public MentorObject(
+      String nameInput,
+      String gitHubIDInput,
+      String descriptionInput,
+      ArrayList<String> tagInput,
+      ArrayList<String> projectIDInputs) {
+    super(nameInput, gitHubIDInput, descriptionInput, tagInput);
+    projectIDs = projectIDInputs;
   }
```
Richie78321, line 34: This is a lot of constructors, do you think we could achieve the same effect by passing `null` to various constructor arguments and handling those accordingly? For example:
```java
public MentorObject(
      String nameInput,
      String gitHubIDInput,
      String descriptionInput,
      ArrayList<String> tagInput,
      ArrayList<String> projectIDInputs) {
    if (tagInput == null) {
      tagInput = new ArrayList<String>();
    }
    [...]
    super(nameInput, gitHubIDInput, descriptionInput, tagInput);
    projectIDs = projectIDInputs;
  }
```

Open to discussion on this.

src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -1,19 +1,43 @@
...
 
-  public String getName() {
-    return name;
+  public ArrayList<String> getProjects() {
```
Richie78321, line 36:
```suggestion
  public ArrayList<String> getProjectIds() {
```
nit: this could be more reflective of what is being returned. In the case that you go with my suggestion above, you can update this too.

src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -1,19 +1,43 @@
...
+    return projectIDs;
```
Richie78321, line 37: This is super nit but I thought it was interesting. If the plan is to mutate the list through the functions provided like `addProject` below, it might be worth considering returning an immutable list here:
```java
return Collections.unmodifiableList(projectIDs);
```

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -14,9 +14,24 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     ArrayList<MentorObject> mentors = new ArrayList<MentorObject>();
-    mentors.add(new MentorObject("Obi", "Obi is awesome."));
-    mentors.add(new MentorObject("Sami", "Sami is fun."));
-    mentors.add(new MentorObject("Richi", "Richi is cool."));
+    ArrayList<String> projects = new ArrayList<String>();
+    projects.add("OpenSesame");
```
Richie78321, line 18: I know this is mostly for demonstration until we get datastore up and running, but I just wanted to make sure that we were in agreement that project IDs will be the datastore-generated IDs and will not be the names of the projects.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -14,9 +14,24 @@
...
+    mentors.add(new MentorObject("Obi", "Obinnabii", "Obi is awesome.", ObiSkills, projects));
+
+    ArrayList<String> SamiSkills = new ArrayList<String>();
+    SamiSkills.add("Stone carver");
+    SamiSkills.add("Bootstrap convert");
+    mentors.add(new MentorObject("Sami", "Sami-2000", "Sami is fun.", SamiSkills, projects));
+
+    ArrayList<String> RichiSkills = new ArrayList<String>();
+    RichiSkills.add("Minecraft boss");
+    RichiSkills.add("React wizard");
```
Richie78321, line 32: loll

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -14,9 +14,24 @@
...
+    mentors.add(new MentorObject("Richi", "Richie78321", "Richi is cool.", RichiSkills, projects));
+
     String jsonMentors = new Gson().toJson(mentors);
```
Richie78321, line 35: nit: Feel free to instantiate this Gson object only once. From the [user guide](https://github.com/google/gson/blob/master/UserGuide.md#TOC-Gson-Performance-and-Scalability):
> The Gson instance does not maintain any state while invoking Json operations. So, you are free to reuse the same object for multiple Json serialization and deserialization operations.

src/main/java/com/google/opensesame/servlets/PersonObject.java
```diff
@@ -0,0 +1,42 @@
...
+    this(nameInput, gitHubIDInput, descriptionInput, new ArrayList<String>());
+  }
+
+  public PersonObject(
+      String nameInput, String gitHubIDInput, String descriptionInput, ArrayList<String> tagInput) {
+    name = nameInput;
+    gitHubID = gitHubIDInput;
+    description = descriptionInput;
+    interestTags = tagInput;
+  }
```
Richie78321, line 25: Same comment as above

src/main/java/com/google/opensesame/servlets/PersonObject.java
```diff
@@ -0,0 +1,42 @@
...
+
+  public String getDescription() {
+    return description;
+  }
+
+  public ArrayList<String> getTags() {
+    return interestTags;
+  }
+
+  public String getGitID() {
```
Richie78321, line 39:
```suggestion
  public String getGitHubID() {
```

src/main/java/com/google/opensesame/servlets/MentorObject.java
Richie78321: Also, something else that I am now thinking of after having looked at the `PersonObject` class is that both the `MentorObject` and future `MenteeObject` class could potentially be more cleanly implemented if the `PersonObject` was stored as a property within these classes instead of as their parent class. Then the `PersonObject` could be passed through the constructor instead of having to pass in every individual property required to construct the `PersonObject`. Is there something you were specifically looking to implement by taking the inheritance approach?

Richie78321: One other thing to consider is that it seems like Gson has some trouble working with polymorphism out of the box. It would most likely require some extra configuration on our end to get serialization to work: https://stackoverflow.com/questions/5800433/polymorphism-with-gson

src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -1,19 +1,43 @@
...
-  private String description;
+import java.util.ArrayList;
 
-  public MentorObject(String nameInput, String descriptionInput) {
-    name = nameInput;
-    description = descriptionInput;
+public class MentorObject extends PersonObject {
+  private ArrayList<String> projectIDs;
+
+  public MentorObject(String nameInput, String gitHubIDInput) {
```
josephdhurst, line 8: In general avoid using `Input` in variable names. Usually the word `input` doesn't tell me any additional information about what the variable will be used for or the information it is holding.

```suggestion
  public MentorObject(String name, String gitHubID) {
```

josephdhurst: I'm only commenting on this method because I don't want to repeat myself, but this advice applies to all methods in this file, not just this one.

src/main/java/com/google/opensesame/servlets/MentorObject.java
josephdhurst: It's best to keep variable names as short as possible while still being descriptive. I think that `projectIDs` is fine as long as there is only one type of project. If there are multiple types of projects and mentoring projects are just one type then `mentoringProjectIds` would make sense because it would disambiguate between the different possible types of project IDs.

src/main/java/com/google/opensesame/servlets/MentorObject.java
josephdhurst: I agree with Richie that this is a lot of constructors. I think that the builder pattern would work well here. Here is a good stackoverflow post describing the builder pattern: https://stackoverflow.com/questions/222214/managing-constructors-with-many-parameters-in-java/222295#222295.

src/main/java/com/google/opensesame/servlets/PersonObject.java
josephdhurst: Also look at my comment above about using the builder pattern to reduce the number of constructors.

src/main/java/com/google/opensesame/servlets/PersonObject.java
```diff
@@ -0,0 +1,42 @@
...
+
+import java.util.ArrayList;
+
+public class PersonObject {
+  private String name;
+  private String description;
+  private ArrayList<String> interestTags;
+  private String gitHubID;
+
+  public PersonObject(String nameInput, String gitHubIDInput) {
```
josephdhurst, line 11: See my comment in MentorObject.java about using `input` in variable names.

src/main/java/com/google/opensesame/servlets/MentorObject.java
Richie78321: Yeah that makes sense. Looking again projectIDs works just fine.

Sami-2000: Thanks for the feedback! It's always helpful to learn more about naming conventions.

src/main/java/com/google/opensesame/servlets/MentorObject.java
Sami-2000: @Richie78321 I believe after our Pod Sunk discussion yesterday, we are now all on the same page about inheritance and class structure.

Sami-2000: Thank you Joe! I will try out that approach.

src/main/java/com/google/opensesame/servlets/MentorObject.java
Sami-2000: Thank you!

src/main/java/com/google/opensesame/servlets/MentorObject.java
Sami-2000: Ooh sounds cool!

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Sami-2000: Quick clarifying Q: Where else do I instantiate a Gson object? (I only see a Gson object declaration on line 36)

Richie78321: You can instantiate it once when the class `MentorServlet` is instantiated like this (not perfect but you get the point): 
```java
class MentorServlet {
  private Gson gson = new Gson();
  [...]
}
That way you are not recreating the Gson object every time a get request is made.

PR #38, opened 2020-07-01
Add mentors page to navbar and add navbar to mentors page.

Sami-2000:
Add updated navbar to mentors page and fix small typo in mentors.js.

### DIFF
--- a/src/main/webapp/mentors.js
+++ b/src/main/webapp/mentors.js
@@ -2,10 +2,10 @@
  * Get mentors from the mentor servlet.
  */
 function getMentors() {
-  console.log("entering get mentors function/n");
+  console.log('entering get mentors function/n');
   let url = new URL('/mentors', location.protocol + '//' + location.hostname);
 
-  fetch(url).then(errorHandling(response)).then(response => response.json()).then(mentors => {
+  fetch(url).then(errorHandling).then(response => response.json()).then(mentors => {
     console.log(mentors);
     const mentorsContainer = document.getElementById('mentors-container');
     mentorsContainer.innerHTML = '';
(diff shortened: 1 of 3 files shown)


src/main/webapp/mentors.js
```diff
@@ -2,10 +2,10 @@
  * Get mentors from the mentor servlet.
  */
 function getMentors() {
-  console.log("entering get mentors function/n");
+  console.log('entering get mentors function/n');
   let url = new URL('/mentors', location.protocol + '//' + location.hostname);
```
Richie78321, line 6: Q: Are you using the URL object here for a particular reason? I'm not sure how this differs from
```javascript
fetch('/mentors')
```
Also, when I ran this code in the demo branch, this wasn't working. It only worked when I added the endpoint directly to the fetch call as shown in the snippet above.

src/main/webapp/mentors.js
```diff
@@ -2,10 +2,10 @@
...
 
-  fetch(url).then(errorHandling(response)).then(response => response.json()).then(mentors => {
+  fetch(url).then(errorHandling).then(response => response.json()).then(mentors => {
```
Richie78321, line 8: This looks good! However, you need to add a catch block at the end of this to catch the errors you're throwing and the errors that can be thrown from the fetch call.

Sami-2000: Thank you! I will add the catch block.

src/main/webapp/mentors.js
Sami-2000: I was using a url object in case we want to add search params later. And, just to clarify, it only ran for you when you used '/mentors' instead of the URL object?

Richie78321: Okay that makes sense for using URL. And yes, that clarification is correct. Let's look into it during our sync.

PR #39, opened 2020-07-01
Add tests to the GitHubGetter

Richie78321:
Adds two tests to the GitHub getter, and restructures the code to make it more easily testable.

Fun fact, I actually already caught a bug with these tests! :)


Richie78321:
Oh lord the commits... something is off here.

PR #40, opened 2020-07-01
Add tests to the GitHubGetter

Richie78321:
Adds two tests to the GitHub getter, and restructures the code to make it more easily testable.

Fun fact, I actually already caught a bug with these tests! :)

### DIFF
--- a/src/main/java/com/google/opensesame/github/GitHubGetter.java
+++ b/src/main/java/com/google/opensesame/github/GitHubGetter.java
@@ -12,7 +12,7 @@ public final class GitHubGetter {
    * available in the system environment variables, the builder will connect to the GitHub API using
    * those. Otherwise, it will connect anonymously.
    *
-   * <p>In AppEngine, the credentials can be set in appengine-web.xml
+   * <p>In AppEngine, the credentials can be set in appengine-web.xml.
    *
    * @return Returns the GitHub API interface.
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/opensesame/github/GitHubGetter.java
```diff
@@ -8,7 +8,11 @@
   private static GitHub gitHub = null;
 
   /**
-   * Gets the GitHub API interface or builds one if there is not one available.
+   * Gets the GitHub API interface or builds one if there is not one available. If credentials are
+   * available in the system environment variables, the builder will connect to the GitHub API using
+   * those. Otherwise, it will connect anonymously.
+   *
+   * <p>In AppEngine, the credentials can be set in appengine-web.xml.
```
billyeh, line 15:
```suggestion
   * In AppEngine, the credentials can be set in appengine-web.xml.
```

Richie78321: Yeah I agree with this change. For some reason this was being enforced by the Java linter.. Although it seems to have stopped now? Hm.

Richie78321: Yeah the Java linter added it back. Any thoughts on why that might happen? Seems odd.

PR #43, opened 2020-07-06
Add a test for associating a project with a mentor object.

Sami-2000:
One super basic test of the addProject function for the mentor object. (hopefully this function in the future may be covered by a more complicated test) Follow the file organization and format from the starter project.

### DIFF
--- a/src/test/java/com/google/opensesame/MentorTest.java
+++ b/src/test/java/com/google/opensesame/MentorTest.java
@@ -8,22 +8,22 @@ import org.junit.runners.JUnit4;
 
 @RunWith(JUnit4.class)
 public class PersonTests {
-  private static final String MENTOR_1_NAME = "Mentor 1";
-  private static final String MENTOR_1_GTHUB = "Mentor_1_GitHub_ID";
-  private static final String PROJECT_1 = "Project 1";
+  private static final String MENTOR_NAME = "Mentor";
+  private static final String MENTOR_GITHUB_ID = "Mentor_GitHub_ID";
+  private static final String PROJECT_ID = "Project_ID";
 
   @Test
   public void addProjects() {
-    MentorObject mentor1 = new MentorBuilder()
-        .name(MENTOR_1_NAME)
-        .gitHubID(MENTOR_1_GITHUB)
+    MentorObject mentor = new MentorBuilder()
+        .name(MENTOR_NAME)
+        .gitHubID(MENTOR_GITHUB_ID)
         .buildMentor();
-    mentor1.addProject(PROJECT_1);
+    mentor.addProject(PROJECT_ID);
     
     List<String> expected = new List<String>;
-    expected.add(PROJECT_1);
+    expected.add(PROJECT_ID);
 
-    List<String> actual = mentor1.getProjectIDs();
+    List<String> actual = mentor.getProjectIDs();
 
     Assert.assertEquals(expected, actual);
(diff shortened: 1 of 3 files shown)


src/test/java/com/google/opensesame/MentorTest.java
```diff
@@ -0,0 +1,30 @@
...
+  private static final String MENTOR_1_GTHUB = "Mentor_1_GitHub_ID";
+  private static final String PROJECT_1 = "Project 1";
+
+  @Test
+  public void addProjects() {
+    MentorObject mentor1 = new MentorBuilder()
+        .name(MENTOR_1_NAME)
+        .gitHubID(MENTOR_1_GITHUB)
+        .buildMentor();
+    mentor1.addProject(PROJECT_1);
```
Obinnabii, line 21: Very nice, I like the mentor builder

src/test/java/com/google/opensesame/MentorTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.io.IOException;
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class PersonTests {
+  private static final String MENTOR_1_NAME = "Mentor 1";
+  private static final String MENTOR_1_GTHUB = "Mentor_1_GitHub_ID";
```
billyeh, line 12:
```suggestion
  private static final String MENTOR_1_GITHUB = "Mentor_1_GitHub_ID";
```

src/test/java/com/google/opensesame/MentorTest.java
```diff
@@ -0,0 +1,30 @@
...
+  private static final String PROJECT_1 = "Project 1";
+
+  @Test
+  public void addProjects() {
+    MentorObject mentor1 = new MentorBuilder()
```
billyeh, line 17: As there's not currently a `mentor2` I suggest naming this variable just `mentor`.

src/test/java/com/google/opensesame/MentorTest.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.List;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+
+@RunWith(JUnit4.class)
+public class PersonTests {
+  private static final String MENTOR_1_NAME = "Mentor 1";
+  private static final String MENTOR_1_GTHUB = "Mentor_1_GitHub_ID";
+  private static final String PROJECT_1 = "Project 1";
```
billyeh, line 13: For these variable names I suggest something like `MENTOR_NAME`, `MENTOR_GITHUB_ID`, and `PROJECT_ID` or such. Reasoning: there's no `MENTOR_2_NAME` right now so there's no need to specify `1`. If we did add a second mentor, we could just make an array and then they'd be indexed in the array and we still wouldn't need `1`, `2`, etc.

And it'd be good to add `_ID` to the ID variables so it's more obvious what they represent.

Sami-2000: Thank you!

Richie78321:
@Sami-2000 It seems like this test leads to a compilation error which I am now receiving when working from the master branch. Please remember to test your code before you submit it.

PR #47, opened 2020-07-06
Fix JSX relative imports

Richie78321:
JSX-preprocessed files will now output to the `js-build` directory. Previously, each individual `react` directory would contain a subdirectory `jsx-preprocessed` that contained the preprocessed versions of the files. This made relative imports difficult, as the relative path between a preprocessed file and a non-react file would change when the file was moved to the `jsx-preprocessed` subdirectory.

This has now been resolved by duplicating the entire `js` directory into the `js-build` directory, so relative imports will not break.
This also has the side effect of now only requiring one call to Babel! No more messy `package.json`!

Because of the directory change, I also had to update the script tags in various html files.


josephdhurst:
> LGTM! I like this a lot!

josephdhurst:
> LGTM! I like this a lot!

I agree with Obi, this seems like a great improvement. Also, Obi remember to approve this PR if you think it's ready to merge. Right now it looks like you haven't approved it yet.

PR #48, opened 2020-07-06
Create User Servlet

Obinnabii:
- Add a Servlet to get a specific user
- Add a JS module to get a specific user

### DIFF
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -11,10 +11,12 @@ import javax.servlet.http.HttpServletResponse;
 
 @WebServlet("/user")
 public class UserServlet extends HttpServlet {
-
+  
+  //Dummy list of people @TODO deprecate this
   private List<PersonObject> people;
 
   @Override 
+  //Instantiate the dummy list of people
   public void init(){
     people = new ArrayList<PersonObject>();
     ArrayList<String> projects = new ArrayList<String>();
@@ -61,6 +63,7 @@ public class UserServlet extends HttpServlet {
   }
 
   @Override
+  //get a specific user. return null if not found.
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {    
     String userString = request.getParameter("user");
 
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -63,7 +63,7 @@ public class UserServlet extends HttpServlet {
   }
 
   @Override
-  // get a specific user. return null if not found.
+  // get a specific user. return null if not found. TODO: User Validation
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String userString = request.getParameter("user");
 
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -1,0 +1,64 @@
+package com.google.opensesame.servlets;
+
+import com.google.gson.Gson;
+import java.io.IOException;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/user")
+public class UserServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {    
+    String userString = request.getParameter("user");
+    ArrayList<PersonObject> people = new ArrayList<PersonObject>();
+    ArrayList<String> projects = new ArrayList<String>();
+    projects.add("OpenSesame");
+
+    ArrayList<String> ObiSkills = new ArrayList<String>();
+    ObiSkills.add("Meme god");
+    ObiSkills.add("HTML wrangler");
+    PersonObject Obi =
+        new PersonBuilder()
+            .name("Obi")
+            .gitHubID("Obinnabii")
+            .description("Obi is awesome.")
+            .interestTags(ObiSkills)
+            .projectIDs(projects)
+            .buildMentor();
+    mentors.add(Obi);
+
+    ArrayList<String> SamiSkills = new ArrayList<String>();
+    SamiSkills.add("Stone carver");
+    SamiSkills.add("Bootstrap convert");
+    MentorObject Sami =
+        new MentorBuilder()
+            .name("Sami")
+            .gitHubID("Sami-2000")
+            .description("Sami is fun.")
+            .interestTags(SamiSkills)
+            .projectIDs(projects)
+            .buildMentor();
+    mentors.add(Sami);
+
+    ArrayList<String> RichiSkills = new ArrayList<String>();
+    RichiSkills.add("Minecraft boss");
+    RichiSkills.add("React wizard");
+    MentorObject Richi =
+        new MentorBuilder()
+            .name("Richi")
+            .gitHubID("Richie78321")
+            .description("Richi is cool.")
+            .interestTags(RichiSkills)
+            .projectIDs(projects)
+            .buildMentor();
+    mentors.add(Richi);
+
+    String jsonMentors = new Gson().toJson(mentors);
+    response.setContentType("application/json;");
+    response.getWriter().println(jsonMentors);
+  }
+}
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -72,9 +72,9 @@ public class UserServlet extends HttpServlet {
             .filter(person -> userString.equals(person.getGitHubID()))
             .findAny()
             .orElse(null);
-    if (result == null){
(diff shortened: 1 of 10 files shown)


src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -0,0 +1,80 @@
...
+            .interestTags(RichiSkills)
+            .projectIDs(projects)
+            .buildPerson();
+    people.add(Richi);
+  }
+
+  @Override
+  // get a specific user. return null if not found.
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    String userString = request.getParameter("user");
```
Richie78321, line 68: Consider validation on this. At this point there should at least be logic to determine if this value is null (meaning the client didn't supply it), in which case you should respond with an error or make a more general query depending on how you intended this to be designed. In the case that you respond with an error, please take a look at [the fetch handler](https://github.com/googleinterns/step22-2020/blob/projects-preview-frontend/src/main/webapp/js/fetch_handler.js) to see how you should be formatting the JSON that is sent in the body of the response. Take a look at [this](https://www.restapitutorial.com/httpstatuscodes.html) for more info about error codes. Error 400 is going to be your best bet for when the user hasn't made a valid request.

In the future when datastore is implemented, you will be building a datastore key from the client-provided ID string. Adding a try-catch block around the [KeyFactory](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/KeyFactory.html#stringtokey) when the ID is being parsed will allow you to determine if the ID is valid and return an error response if it isn't.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -0,0 +1,80 @@
...
+
+    PersonObject result =
+        people.stream()
+            .filter(person -> userString.equals(person.getName()))
+            .findAny()
+            .orElse(null);
```
Richie78321, line 74: Is this intended to query by name? Or is querying by user ID intended here? If the request is querying by name you should definitely be coding for multiple results as there could be multiple users with the same name.

src/main/webapp/user.js
```diff
@@ -0,0 +1,26 @@
...
+function getUser(user) {
+  console.log('entering get user function/n');
+  fetch('/user?user=' + user)
+    .then(errorHandling).then((response) => response.json())
+    .then((user) => {
+      return user;
+    })
+    .catch((error) => {
+      console.log(error);
+    });
```
Richie78321, line 15: Move this over to the [standardized fetch error handler](https://github.com/googleinterns/step22-2020/blob/projects-preview-frontend/src/main/webapp/js/fetch_handler.js) in the future. Totally fine for now, especially because the error handler isn't even merged with master yet lol.

src/main/webapp/user.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Get the user @param user from the user servlet.
+ * @param {String} user
```
Richie78321, line 3: nit
```suggestion
 * @param {string} user
```
https://jsdoc.app/tags-type.html

src/main/webapp/user.js
```diff
@@ -0,0 +1,26 @@
...
+ * @return a user object
+ */
+function getUser(user) {
+  console.log('entering get user function/n');
+  fetch('/user?user=' + user)
```
Richie78321, line 8: Consider using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) in the future if you plan on adding more parameters.

src/main/webapp/user.js
```diff
@@ -0,0 +1,26 @@
...
+    .then(errorHandling).then((response) => response.json())
+    .then((user) => {
+      return user;
+    })
```
Richie78321, line 12: Same as above. If the intention is for this to query a name instead of an ID (which is unique to each user), you should code this for multiple results.

src/main/webapp/user.js
```diff
@@ -0,0 +1,26 @@
...
+    .catch((error) => {
+      console.log(error);
+    });
+}
+
+/**
+ * Basic error handling checks if fetch results are 'ok.'
+ */
```
Richie78321, line 20: Fix JSDoc according to ESLint pls

src/main/webapp/user.js
Obinnabii: Interesting, I thought String was a custom class like in Java haha

Obinnabii:
```suggestion
 * @param {string} user
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: The user isn't making requests here. this is completely automated and it returns null if an invalid sting is given.

src/main/webapp/user.js
Obinnabii: I changed this in a future PR

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: but interesting, I'll look into the keyfactory stuff.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -0,0 +1,80 @@
...
+
+  // Dummy list of people @TODO deprecate this
+  private List<PersonObject> people;
+
+  @Override
+  // Instantiate the dummy list of people
+  public void init() {
+    people = new ArrayList<PersonObject>();
+    ArrayList<String> projects = new ArrayList<String>();
+    projects.add("OpenSesame");
```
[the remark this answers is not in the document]
Obinnabii: these are saved as projectIDs in the projectIDs section of the user object.

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: It isn't, I was following the mentorServlet convention. Good catch though!

src/main/java/com/google/opensesame/servlets/UserServlet.java
Richie78321: I understand that this is not intended for public use, but the fact that this endpoint is publicly accessible means that we need to add validation in anticipation of faulty requests. Also, this endpoint is intended for "public use" within our intern pod, which means that there should be easy-to-diagnose error codes in the case that someone uses this improperly.

Obinnabii: I'll add it then

Richie78321: Lmk when you've added this validation stuff and I'll re-review.

billyeh: It's good to have an idea of what the code should look like in its final form, and to have agreement on that. 

But a PR not reaching that final, ideal form isn't a good reason to withhold approval. Reasoning: will our code ever reach its final, ideal form? Probably not. So that doesn't mean we should never commit any code until we reach that state. We need to make incremental progress. It's totally fine to ask for/put a TODO here saying "TODO Add validation" and address it later.

We also need to prioritize in the project, given the time constraints. This is why it would be nice to have these things estimated in the design doc, and how long it would take. Is validation necessary for the MVP? Final presentation? Is it something we should include in "future work", after the internship?

Let me know if that makes sense!

Richie78321: This makes a lot of sense and is solid advice. I'll keep this in mind moving forward, thanks!

Richie78321: And @Obinnabii, I'll approve this now, feel free to add a TODO here about validation.

Richie78321: And @billyeh just to make sure I understand 100% I'm going to put it in my own words: you're saying that PRs are meant for verifying that the functionality that has been implemented has been implemented properly, not to request additional functionality that could come in a future PR. This makes sense!

billyeh: Yeah, nice. Not a huge deal, just something I wanted to point out again. 

We went through writing the design doc etc. to get agreement on what features to include in the capstone project, and how to implement them. As long as a PR is making progress toward that goal/in that way, it should be fine.

It is a really good idea to keep in mind things like validation, styling etc. But I didn't see that as a high priority/feature in the design doc, and it's probably not as crucial as getting a working product currently.

josephdhurst: Even if you don't plan on adding validation right now, add a TODO to help yourself keep track of the possible improvements to the code.

Obinnabii: I added it below

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -0,0 +1,92 @@
...
+            .gitHubID("Richie78321")
+            .description("Richi is cool.")
+            .interestTags(RichiSkills)
+            .projectIDs(projectIDs)
+            .buildPerson();
+    people.add(Richi);
+  }
+
+  @Override
+  // get a specific user. return null if not found.
```
Obinnabii, line 66:
```suggestion
  // get a specific user. return null if not found. TODO: User Validation
```

PR #49, opened 2020-07-06
Add detail to mentors page.

Sami-2000:
Bring the frontend of the mentors page to MVP level by adding a github link, interest tags, and project cards.


src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -57,6 +57,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
     mentors.add(Richi);
 
     String jsonMentors = new Gson().toJson(mentors);
+    System.out.println("HELLLLLLLOOOOO" + jsonMentors);
```
Richie78321, line 60: Is this intended to be put into master? Lol

Richie78321, line 60: Is this intended to be put into master? Lol

Sami-2000: Omg no! Just a debugging comment xo

src/main/webapp/mentors.js
```diff
@@ -51,12 +59,72 @@ function createMentorElement(name, description) {
...
+  projectsDiv.className = 'row p-3';
+  for (const projectID of projectIDs) {
+    let projElement = document.createElement('div');
+    projElement.className = 'card container card-holder col-12' +
+        ' text-center project-card p-3 m-3';
+    let cardTitleElement = document.createElement('h4');
+    cardTitleElement.className = 'card-title dark-emph';
+    cardTitleElement.innerText = projectID;
+    projElement.appendChild(cardTitleElement);
+    projectsDiv.append(projElement);
```
Richie78321, line 127: nit: Don't abbreviate for readability risk
```suggestion
    let projectElement= document.createElement('div');
    projectElement.className = 'card container card-holder col-12' +
        ' text-center project-card p-3 m-3';
    let cardTitleElement = document.createElement('h4');
    cardTitleElement.className = 'card-title dark-emph';
    cardTitleElement.innerText = projectID;
    projectElement.appendChild(cardTitleElement);
    projectsDiv.append(projectElement);
```

src/main/webapp/mentors.js
```diff
@@ -51,12 +59,72 @@ function createMentorElement(name, description) {
...
+function createProjectsRow(projectIDs) {
+  let projectsDiv = document.createElement('div');
+  projectsDiv.className = 'row p-3';
+  for (const projectID of projectIDs) {
+    let projElement = document.createElement('div');
+    projElement.className = 'card container card-holder col-12' +
+        ' text-center project-card p-3 m-3';
+    let cardTitleElement = document.createElement('h4');
+    cardTitleElement.className = 'card-title dark-emph';
+    cardTitleElement.innerText = projectID;
```
Richie78321, line 125: Same issue as above. The project ID is a numeric identifier and is not readable. I think the assumption here is that the `projectID` is the name of the GitHub repository, which it is not.

src/main/webapp/mentors.js
```diff
@@ -1,18 +1,22 @@
...
-    console.log(mentors);
     const mentorsContainer = document.getElementById('mentors-container');
     mentorsContainer.innerHTML = '';
     for (const mentor of mentors) {
-      mentorsContainer
-          .appendChild(createMentorElement(mentor.name, mentor.description));
+      for (const propName in mentor) {
+        const propVal = mentor[propName];
+        console.log(propName, propVal);
+      }
```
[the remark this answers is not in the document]
Sami-2000: It is slightly different in that it reports the properties explicitly (instead of giving an object which you can then inspect). However, if you think it is simpler/better, I can switch to console.log(mentor) and not lose any information.

src/main/webapp/mentors.js
```diff
@@ -1,18 +1,22 @@
...
+      mentorsContainer.appendChild(createMentorElement(mentor.name, 
+          mentor.description, mentor.gitHubID, mentor.interestTags, mentor.projectIDs));
```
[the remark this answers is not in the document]
Sami-2000: Thank you!

src/main/webapp/mentors.js
Sami-2000: I'm glad we had the chance to talk about this ID comment in person :) Like we agreed, eventually, when we have concretized how we will load to and draw from datastore, there will be an in-between step here to fetch the readable project name, mentor name (and most of the personal attributes for the matter) but I am glad we decided to leave the readable strings for this iteration.

src/main/webapp/mentors.js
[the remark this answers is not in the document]
Richie78321: Okay, thanks for the clarification! Totally up to you if you want to keep it or not.

src/main/webapp/mentors.js
Richie78321: Yeah as long as we're both on the same page we're fine. This'll work for now!

src/main/webapp/mentors.js
```diff
@@ -51,12 +59,72 @@ function createMentorElement(name, description) {
...
+function createGitHubLink(gitHubID) {
+  let gitSrc = 'https://github.com/';
+  const gitLink = gitSrc.concat(gitHubID);
+  let userGithubButton = document.createElement('a');
+  userGithubButton.innerText = 'GitHub Profile';
+  userGithubButton.className = 'btn btn-primary';
+  userGithubButton.role = 'button';
+  userGithubButton.href = gitLink;
+  return userGithubButton;
+}
```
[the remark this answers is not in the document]
Richie78321: Now that I know we're on the same page, I'm not as worried about this and we can iterate and a future PR.

src/main/webapp/mentors.js
```diff
@@ -48,15 +50,75 @@ function createMentorElement(name, description) {
...
+
+/**
+ * Creates a block of interest tags.
+ * @param {String[]} interestTags
+ * @return {HTMLElement} tagDiv
+ */
+function createTagRow(interestTags) {
```
josephdhurst, line 93: nit: call this `createTagDiv` because `createTagRow` makes me think that this is creating a `<tr>` object.

src/main/webapp/mentors.js
```diff
@@ -48,15 +50,75 @@ function createMentorElement(name, description) {
...
+  }
+  return tagDiv;
+}
+
+/**
+ * Creates a block of project cards.
+ * @param {String[]} projectIDs
+ * @return {HTMLElement} projectsDiv
+ */
+function createProjectsRow(projectIDs) {
```
josephdhurst, line 110: Same comment as above. Only use row when referring to `<tr>` objects.

src/main/webapp/mentors.js
Sami-2000: Thank you!

PR #50, opened 2020-07-06
Add fetch support on projects preview frontend

Richie78321:
Adds fetch support on the projects preview frontend.

* Added a standardized error formatter for fetch requests being made to the Open Sesame API
* Restructured the frontend slightly to match the changes made on the backend

I commented the fetch error formatter a lot in anticipation of it being used all over the place. Please let me know if these comments aren't good enough, as I'm worried they might not make sense to someone who has done less research on this.


src/main/webapp/js/fetch_handler.js
```diff
@@ -0,0 +1,100 @@
...
+ * error from the Fetch API, a server error from the Open Sesame API, and a
+ * defined API error from the Open Sesame API.
+ *
+ * All error responses from the Open Sesame API should contain a JSON body
+ * formatted to match the ErrorResponse object here. {@link ErrorResponse}.
+ */
+/**
+ * @typedef ErrorResponse
+ * @property {string} error
+ * @property {number} [statusCode]
```
Sami-2000, line 13: Just a Q: What does it mean that statusCode is in brackets?

src/main/webapp/js/projects/react/components/ProjectsSearch.js
```diff
@@ -0,0 +1,61 @@
...
+
+  /**
+   * Renders the list of project previews or a loading notification if the data
+   * is still being loaded.
+   * @return {React.Component}
+   */
+  render() {
+    if (this.state.isFetching) {
+      return (
+        <h1 className="text-primary text-center mt-2">Loading projects...</h1>
```
Sami-2000, line 55: Just a Q: If this message pops up, at what point is it removed? (will re-rendering automatically remove it?)

Richie78321:
> Also, just a thought, but if the server always responds well, how can we test these server error response functions? (looking at the jest site, I don't see clear ways to mess with the HTTP stack. Seems like API nock may allow you to test this more easily? https://github.com/nock/nock)

This is a good point. I should definitely add some testing here. I'll address this in an upcoming PR.

src/main/webapp/js/fetch_handler.js
```diff
@@ -0,0 +1,100 @@
...
+ * @param {string} htmlString
+ * @return {string} Returns the error text.
+ */
+function extractErrorTextFromHTML(htmlString) {
+  const responseDOM = new DOMParser().parseFromString(htmlString, 'text/html');
+  const body = responseDOM.getElementsByTagName('BODY')[0];
+
+  return Array.from(body.childNodes)
+      .map((child) => child.textContent).join(' ');
+}
```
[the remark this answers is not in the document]
Richie78321: I just looked through the Google JS style guide and it seems like there isn't an explicit convention. However, GitHub displays an error when there isn't a newline at the end of a file, so I add it to comply with that.

src/main/webapp/js/projects/react/components/ProjectPreview.js
```diff
@@ -9,13 +9,19 @@ import {projectPreviewType} from '../prop_types.js';
 export default function ProjectPreview(props) {
   const projectPreview = props.projectPreview;
```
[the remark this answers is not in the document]
Richie78321: Very nit but yeah I agree. I updated the file.

src/main/webapp/js/fetch_handler.js
Richie78321: It indicates that the `statusCode` property is optional and will potentially not be present in some `ErrorResponse` objects. Good eye!

You can read about it [here](https://jsdoc.app/tags-type.html) (Ctrl+F "optional" to find it).

src/main/webapp/js/projects/react/components/ProjectsSearch.js
Richie78321: Yes. Changing the state of a component with `this.setState()` will automatically trigger a re-render, at which point React will replace whatever is present in the DOM from the previous render with what is returned from the render function here.

src/main/webapp/js/projects/react/components/ProjectsSearch.js
```diff
@@ -0,0 +1,61 @@
...
+          `Error ${errorResponse.statusCode}: ${errorResponse.error}`);
+    });
+  }
+
+  /**
+   * Renders the list of project previews or a loading notification if the data
+   * is still being loaded.
+   * @return {React.Component}
+   */
+  render() {
```
josephdhurst, line 52: I think that this function should be named something like `loadProjectPreviews`. `render` is used to describe logic that actually displays images on the screen (https://en.wikipedia.org/wiki/Rendering_(computer_graphics)). It seems like this function just returns data and doesn't handle any of the graphics display logic.

Richie78321: You make a good point, however [this is a function associated with the React.Component class](https://reactjs.org/docs/react-component.html#render) that is being extended here. I'm going to add the @override tag as per the [style guide](https://google.github.io/styleguide/jsguide.html#jsdoc-method-and-function-comments) to make this more clear.

PR #54, opened 2020-07-07
Switch 'location' variable to 'window.location'

Sami-2000:


src/main/webapp/mentors.js
```diff
@@ -3,7 +3,7 @@
  */
 function getMentors() {
   console.log('entering get mentors function/n');
-  let url = new URL('/mentors', location.protocol + '//' + location.hostname);
+  let url = new URL('/mentors', window.location.protocol + '//' + window.location.hostname);
```
Obinnabii: It might be good to look into using [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)

billyeh:
I think this PR doesn't change any files? :) I made the change in 81e924c5fc810a17dd9bbcd6571d5991137360e3 already.

PR #56, opened 2020-07-07
Add basic authentication servlet

Richie78321:
Most of this code is being sourced from [my portfolio project](https://github.com/Richie78321/google-step/blob/master/portfolio/src/main/java/com/google/sps/servlets/AuthServlet.java).

The initial plan was to use OAuth for both authorizing GitHub and authenticating the user on our service, so that the user would only ever have to log in with GitHub. However, an error on the GCP similar to the Week-5 API errors is occurring that is preventing us from using the [Firebase Admin SDK](https://firebase.google.com/docs/admin/setup) on our server.

Our second-best option is this. We are using the deprecated [Users API](https://cloud.google.com/appengine/docs/standard/java/users) for platform authorization, and then using OAuth to link a user with a GitHub profile.

For now the login page will redirect to the home page as dictated by the `AUTH_URL_REDIRECT` constant, but this will most likely change to the user dashboard or similar down the line.


Richie78321:
Removing the internal class as per Joe's tips about avoiding unnecessary overhead.

src/main/java/com/google/opensesame/auth/AuthServlet.java
```diff
@@ -0,0 +1,72 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {    
+    JsonObject responseObject = new JsonObject();
+
+    UserService userService = UserServiceFactory.getUserService();
+    String loginUrl = userService.createLoginURL(AUTH_URL_REDIRECT);
+    String logoutUrl = userService.createLogoutURL(AUTH_URL_REDIRECT);
+
+    responseObject.addProperty("loginUrl", loginUrl);
+    responseObject.addProperty("logoutUrl", logoutUrl);
```
josephdhurst, line 50: In this case storing the login/logout urls in a variable doesn't help readability. The variable names `loginUrl` and `logoutUrl` don't add any information because they have the same names as the properties in the response object.

```suggestion

    responseObject.addProperty("loginUrl", userService.createLoginURL(AUTH_URL_REDIRECT));
    responseObject.addProperty("logoutUrl", userService.createLogoutURL(AUTH_URL_REDIRECT));
```

Richie78321: That makes sense, thanks!

PR #57, opened 2020-07-08
Add React authorization button to navbar

Richie78321:
* Add a React-based authorization button that fetches authorization data from the authorization servlet and updates its appearance to either be a "Log In" button or a "Log Out" button.
* Add this authorization button to the navbar for easy website-wide access.
* Update the main style sheet `style.css` to be more permitting of other buttons other than those on the main home page.

The intended flow is for a new user to sign up with this button, from which they are redirected (by the login page) to their profile, where they can set data about themselves and link their GitHub accounts.

### DIFF
--- a/src/main/webapp/js/common/react/components/AuthButton.js
+++ b/src/main/webapp/js/common/react/components/AuthButton.js
@@ -10,7 +10,7 @@ checkTesting();
 export default class AuthButton extends React.Component {
   /**
    * Create the AuthButton react component.
-   * @param {{}} props 
+   * @param {{}} props
    */
   constructor(props) {
     super(props);
@@ -55,7 +55,7 @@ export default class AuthButton extends React.Component {
    * @return {React.Component}
    * @override
    */
-  render() { 
+  render() {
     let classList = 'btn btn-primary';
     if (this.state.isFetching) {
       // The button is disabled while the authorization data is being loaded.
--- a/src/main/webapp/js/common/react/components/AuthButton.js
+++ b/src/main/webapp/js/common/react/components/AuthButton.js
@@ -25,7 +25,7 @@ export default class AuthButton extends React.Component {
    * @override
    */
   componentDidMount() {
-     const fetchRequest = standardizeFetchErrors(fetch('/auth'),
+    const fetchRequest = standardizeFetchErrors(fetch('/auth'),
         'Failed to communicate with the server, please try again later.',
         'Encountered a server error, please try again later.');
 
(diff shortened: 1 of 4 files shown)


Richie78321:
Waiting for #50 to be merged into master to clean up the commits here.

src/main/webapp/js/common/react/components/AuthButton.js
```diff
@@ -0,0 +1,83 @@
...
+    this.state = {
+      isFetching: true,
+    };
+  }
+
+  /**
+   * Loads user authorization data after React has initialized the component.
+   * @override
+   */
+  componentDidMount() {
```
Sami-2000, line 27: Q: Does this function only call another function? (if so, could you just move the body of 'fetchAuthorizationStatus' here? or will you need that function other places)

src/main/webapp/style.css
```diff
@@ -11,10 +11,25 @@ a:hover,
...
   color: white;
-  width: 70%;
+}
+
+.btn.disabled {
+  background-color: rebeccapurple;
+  color: white;
+}
+
+.btn-shadow:hover {
```
Sami-2000, line 31: Q: To clarify: now Obi and I need to add this class if we want box shadows on hover? (if we need to make this change plz throw it in 3 peas chat when u merge so he and I can change our code accordingly)

src/main/webapp/js/common/react/components/AuthButton.js
```diff
@@ -0,0 +1,83 @@
...
+    });
+  }
+
+  /**
+   * Renders the Login/Logout button. If the authorization data is still
+   * loading, the button is disabled.
+   * @return {React.Component}
+   * @override
+   */
+  render() {
```
Sami-2000, line 58: Trailing spaces seem like the nittiest nit, but I suppose if you push this branch again you may as well remove them so we can get rid of these linting messages.

billyeh: Yes please. Let's get rid of all linter errors before merging anything to master :)

Richie78321: Sgtm! Thanks for the reminder. Will update those now.

src/main/webapp/js/common/react/components/AuthButton.js
Richie78321: Good point. I think this is a remnant of thinking I was going to need it more than once. Will update!

src/main/webapp/style.css
Richie78321: Yeah you will. Sorry for doing this but it'll be nice to have the option of a shadow on some buttons. I'll send out a chat now.

PR #60, opened 2020-07-09
Add projects datastore support

Richie78321:
Adds support for adding and updating projects in the datastore. This functionality is only accessible through `ProjectDatastore` functions and not through a servlet, as this functionality is only used indirectly from a client request. For example, requesting data about a project will go through a separate Projects servlet that will populate response data based on information retrieved from the datastore. That will be coming in a future PR.

* Add the `ProjectDatastore` helper functions for storing and retrieving project entities from storage.
* Add the `ProjectEntity` class for making the creation and updating of project datastore entities more straightforward.
* Updated the name of `ProjectPreview` to `ProjectPreviewData` to more clearly illustrate that it is intended primarily for use in JSON serialization.

The next steps after this are going to be creating the aforementioned projects servlet and the accompanying `ProjectData` class that will be used for populating and serializing more in-depth data about a project than the existing `ProjectPreviewData`. This will ultimately be used for the expanded project view. However, both of these things are non-critical for the MVP.


src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -0,0 +1,116 @@
...
+  public static final String INTERESTED_USER_IDS_KEY = "interestedUserIds";
+
+  /**
+   * Creates a new ProjectEntity from a GitHub repository name.
+   *
+   * @param repositoryName
+   * @return Returns the new ProjectEntity or null if the repository name is invalid.
+   * @throws IOException
+   */
+  public static ProjectEntity fromRepositoryName(String repositoryName) throws IOException {
```
Obinnabii, line 28: Would it be possible to instantiate this with a field for adding a mentor? for now, all projects will pop up with at least one mentor.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -0,0 +1,116 @@
...
+   * @param mentorIds
+   * @param interestedUserIds
+   */
+  public ProjectEntity(
+      String repositoryId, List<String> mentorIds, List<String> interestedUserIds) {
+    this.repositoryId = repositoryId;
+    this.mentorIds = mentorIds;
+    this.interestedUserIds = interestedUserIds;
+  }
+
```
Obinnabii, line 77: Maybe an ```addMentor()``` function if that makes the flow easier.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: Should be doable with
```java
ProjectEntity newProjectEntity = ProjectEntity.fromRepositoryName(repoName);
newProjectEntity.mentorIds.add(newMentor);
```

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -0,0 +1,47 @@
...
+    Entity newProjectEntity = new Entity(PROJECT_KIND, projectEntity.repositoryId);
+    projectEntity.fillPropertyContainer(newProjectEntity);
+
+    return datastore.put(newProjectEntity);
+  }
+
+  /**
+   * Gets a project entity by its repository ID.
+   *
+   * @param repositoryId The repository ID to query for. The repository ID is the must precise way
```
Sami-2000, line 31: Nit: 'most'

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: That's a good point. Both the `mentorIds` and `interestedUserIds` lists are publicly accessible at the moment, but I'm wondering if it'd be more safe to make them both private and only allow for adding and removing. Hmm... What do you think? Also Bill and Joe feel free to chime in here on what you would consider the best practices.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Sami-2000, line 28: Just a Q: If GitHub repository names are not unique, how can we get a unique ID based soley on the name?

Sami-2000: Do you also need an organization name, or GitHub info from the logged-in user for this function?

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
Richie78321: Haha nice catch

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: Answered over chat but will answer here so others can see. The repository name takes the form `user/repo` or `org/repo` and in the short term it is unique. However org names and repo names can change so we only treat it as unique for this short timeframe between when the user supplies the repo URL and when it is converted into the permanent repo ID.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -0,0 +1,47 @@
...
+public class ProjectDatastore {
+  private static DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+  public static final String PROJECT_KIND = "Project";
+
+  /**
+   * Adds a project entity to the datastore, or overwrites a project entity if an entity with the
+   * same repository ID exists.
+   *
+   * @param projectEntity
+   * @return Returns the key associated with the datastore entity;
```
billyeh, line 19:
```suggestion
   * @return Returns the key associated with the datastore entity.
```

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -0,0 +1,116 @@
...
+  public int getNumInterestedUsers() {
+    return this.interestedUserIds.size();
+  }
+
+  /**
+   * Creates a map of properties from this ProjectEntity object.
+   *
+   * @return Returns the map of properties.
+   */
+  public Map<String, Object> createProperties() {
```
billyeh, line 91:
```suggestion
  private Map<String, Object> createProperties() {
```

Does it need to be public? Just wondering, seems this would mainly just be used for `fillPropertyContainer` below.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  /**
+   * Adds the properties of this ProjectEntity object to a PropertyContainer, which will most likely
+   * be a datastore entity, which is a child of PropertyContainer:
+   * https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/Entity
+   *
+   * @param propertyContainer The property container to add this ProjectEntity's properties to.
+   */
+  public void fillPropertyContainer(PropertyContainer propertyContainer) {
```
billyeh, line 109: Just wondering, would it make sense to create and return a `PropertyContainer` from this function, so a user doesn't have to pass it in? 

Kind of a theoretical question since I don't see the function used yet, just wanted to see if you think one way or another is better.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
billyeh: Yeah having an addMentor function might be convenient, maybe not so much here but related to the Datastore? But it's also probably fine just to use the public field and save the whole project entity again for now.

Richie78321: @billyeh You make a good point, however Datastore is designed such that updating an entity [involves saving it back to the Datastore](https://cloud.google.com/appengine/docs/standard/java/datastore/creating-entities#Updating_entities) which is why I thought it was best to create this class as a clean way to interface with the properties of an existing Datastore entity.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: It's used for functions that require the properties from an entity, with a good example being `ProjectPreview.fromProperties(projectEntity.createProperties())` where a project is pulled from the Datastore, and project preview data is prepared to be sent to the client.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: `PropertyContainer` is used because it is the parent class of `Entity` and allows us to set the properties of the entity. However, I wanted to avoid using `Entity` directly here in case I wanted to extend the `PropertyContainer` to create a mock for testing.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
billyeh: Sounds good! nit: might just want to name `addProjectEntity` `putProjectEntity` since it adds or updates.

PR #63, opened 2020-07-09
Add instructions for running eslint locally

billyeh:
Realized we can run this locally and check for errors before pushing to Github, since I think I talked with @Richie78321 about it this morning thought I should contribute the instructions.


billyeh:
Note that the linter errors for Javascript are actual errors :) Let's try to make sure to keep the linter happy before merging to master.

PR #67, opened 2020-07-10
Handle invalid mentor form input.

Sami-2000:
When user inputs something other than a github url, or a github url which the API can't link to a project, print an error message on the mentor form page.

### DIFF
--- a/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
@@ -88,49 +88,60 @@ public class MentorsServlet extends HttpServlet {
 
   //This function prints an error message to the mentor form page if invalid input is found.
   public void formError(HttpServletRequest request, HttpServletResponse response, String error)  throws ServletException, IOException {
-    String message = "<ul><li class='ml-3 display-4 font-weight-bold text-danger'>"   + error + "</li></ul>";
+    System.out.println("form error!");
+    response.setContentType("application/json");
+    response.getWriter().println(new Gson().toJson(error));
+    /*String message = "<ul><li class='ml-3 display-4 font-weight-bold text-danger'>"   + error + "</li></ul>";
     response.setContentType("text/html");
     PrintWriter pw = response.getWriter();
     pw.print(message);
     RequestDispatcher rd = request.getRequestDispatcher("mentor_form.html");
     rd.include(request, response);
-    pw.close();
+    pw.close();*/
     return;
   }
 
   @Override
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws ServletException, IOException {
+    System.out.println("request is " + request);
     String repoUrl = request.getParameter("inputRepo");
-    String urlPattern = "https://github.com/";
+    System.out.println("repo is " + repoUrl);
+    /*String urlPattern = "https://github.com/";
     Pattern regEx = Pattern.compile(urlPattern);
     Matcher inputMatcher = regEx.matcher(repoUrl);
     if (!inputMatcher.find()) {
       //Input is not in the format of a GitHub repo link.
       formError(request, response, "ENTER A VALID REPO LINK");
       return;
-    }
+    }*/
 
     String repoName = repoUrl.replaceFirst("https://github.com/","");
     GitHub gitHub = GitHubGetter.getGitHub();
-    GHRepository inputRepo = gitHub.getRepository(repoName);
-    if (inputRepo == null) {
-      //Repo was not found on GitHub.
+    try { 
+      GHRepository inputRepo = gitHub.getRepository(repoName);
+    } catch(Exception e) {
+      System.out.println("Couldn't find repo\n");
       formError(request, response, "COULDN'T FIND REPO");
       return;
     }
-    //TODO: Send Richie the GHRepository to add to projects database
+
+    //TODO: Send Richie the GHRepository ID and person ID to add to projects database
+    //      Send to a function in the ProjectEntity.java class
 
     String userID = AuthServlet.getAuthorizedUser().getUserId();
     
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    try { 
+    /*try { 
       Entity personEntity = datastore.get(KeyFactory.stringToKey(userID)); 
-    } catch (EntityNotFoundException e) {
+    } catch (Exception e) {
+      formError(request, response, "COULDNT FIND U IN DATASTORE");
+      return;
       //TODO: Handle this exception (which should never occur if datastore is working properly)
-    }
+    }*/
 
     //TODO: Add a mentor entity as a child of the user entity.
 
-    response.sendRedirect("/dashboard.html");
+    response.setContentType("application/json");
+    response.getWriter().println(new Gson().toJson("success"));
   }
 }
--- a/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
@@ -2,9 +2,6 @@ package com.google.opensesame.servlets;
 
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
-import com.google.appengine.api.datastore.Entity;
-import com.google.appengine.api.datastore.EntityNotFoundException;
-import com.google.appengine.api.datastore.KeyFactory;
 import com.google.gson.Gson;
 import com.google.gson.JsonObject;
 import com.google.opensesame.auth.AuthServlet;
@@ -72,8 +69,9 @@ public class MentorsServlet extends HttpServlet {
     response.getWriter().println(jsonMentors);
   }
 
-  //This function sends an error if invalid input is found.
-  public void error(HttpServletResponse response, String errorMessage, int statusCode, String userMessage)  
+  // This function sends an error if invalid input is found.
+  public void error(
+      HttpServletResponse response, String errorMessage, int statusCode, String userMessage)
       throws ServletException, IOException {
     JsonObject responseObject = new JsonObject();
     responseObject.addProperty("message", errorMessage);
@@ -105,7 +103,7 @@ public class MentorsServlet extends HttpServlet {
     String userID = AuthServlet.getAuthorizedUser().getUserId();
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
-    //This section is commented out until we have populated datastore to refer to.
+    // This section is commented out until we have populated datastore to refer to.
     /*try {
       Entity personEntity = datastore.get(KeyFactory.stringToKey(userID));
     } catch (EntityNotFoundException e) {
--- a/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
@@ -6,6 +6,7 @@ import com.google.appengine.api.datastore.Entity;
 import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.gson.Gson;
+import com.google.gson.JsonObject;
 import com.google.opensesame.auth.AuthServlet;
 import com.google.opensesame.github.GitHubGetter;
 import java.io.IOException;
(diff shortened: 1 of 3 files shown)


src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    String message =
+        "<ul><li class='ml-3 display-4 font-weight-bold text-danger'>" + error + "</li></ul>";
+    response.setContentType("text/html");
+    PrintWriter pw = response.getWriter();
+    pw.print(message);
+    RequestDispatcher rd = request.getRequestDispatcher("mentor_form.html");
+    rd.include(request, response);
+    pw.close();
+    return;
+  }
```
Richie78321, line 90: Can we have a call about this? I think it's important that we stick to our convention of JSON error responses, and I think there's any easy fix here. Take a look at the [fetch error handler](https://github.com/googleinterns/step22-2020/blob/master/src/main/webapp/js/fetch_handler.js) for more information on the convention I'm talking about. Instead of returning HTML, you can use JSON to return the error and add it to the HTML on the client side.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      throws ServletException, IOException {
+    String repoUrl = request.getParameter("inputRepo");
+    String urlPattern = "https://github.com/";
+    Pattern regEx = Pattern.compile(urlPattern);
+    Matcher inputMatcher = regEx.matcher(repoUrl);
+    if (!inputMatcher.find()) {
+      // Input is not in the format of a GitHub repo link.
+      formError(request, response, "ENTER A VALID REPO LINK");
+      return;
+    }
```
Richie78321, line 103: Let's also talk about this in a call. This regex string won't be able to catch all invalid repository URLs. For example, if a user inputs their profile URL here (like `https://github.com/Richie78321`) it will still match.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    String repoName = repoUrl.replaceFirst("https://github.com/", "");
+    GitHub gitHub = GitHubGetter.getGitHub();
+    GHRepository inputRepo = gitHub.getRepository(repoName);
+    if (inputRepo == null) {
+      // Repo was not found on GitHub.
+      formError(request, response, "COULDN'T FIND REPO");
+      return;
+    }
+    // TODO: Send Richie the GHRepository to add to projects database
```
Richie78321, line 113: When I implement my projects database function for this I think it'd be best if all of this logic is handled in the projects database function that you mention in the TODO. Keeps this function more clean and allows the code to be reused in other places if necessary. We can also talk about this on a call.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    String userID = AuthServlet.getAuthorizedUser().getUserId();
+
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    try {
+      Entity personEntity = datastore.get(KeyFactory.stringToKey(userID));
+    } catch (EntityNotFoundException e) {
+      // TODO: Handle this exception (which should never occur if datastore is working properly)
+    }
```
Richie78321, line 122: We should also talk about this in a call. I think it would be best if this logic is placed in an AuthServlet helper function, and one is going to be implemented soon.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
+    // TODO: Add a mentor entity as a child of the user entity.
+
+    response.sendRedirect("/dashboard.html");
```
Richie78321, line 126: We should also talk about this in a call.

Richie78321:
Sorry for not being too descriptive with this but I'm thinking if we can get on a call soon it'd be more efficient than typing it all out.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Richie78321: Didn't realize the full context of this as part of a fetch call being made directly from a form. That's my bad. This is totally fine!

Richie78321:
Removed some comments that we didn't get to b/c they didn't make sense after getting more context haha sorry.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -60,4 +74,55 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    pw.close();
+    return;
+  }
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response)
+      throws ServletException, IOException {
+    String repoUrl = request.getParameter("inputRepo");
+    String urlPattern = "https://github.com/";
+    Pattern regEx = Pattern.compile(urlPattern);
```
josephdhurst, line 97: nit: the variable name `urlPattern` doesn't give the reader any additional information they wouldn't be able to quickly figure out from reading `"https://github.com/"`.
```suggestion
    Pattern regEx = Pattern.compile("https://github.com/");
```

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
josephdhurst, line 103: I think that it would be simpler to just rely on the `gitHub.getRepository` function to figure out whether a repository is valid or not. If a url doesn't start with `https://github.com/` I'd assume that the `gitHub.getRepository` function wouldn't be able to find the repo and then the user would get the `COULDN'T FIND REPO` message and they'd be able to figure out what was wrong from that error message.
```suggestion
```

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
josephdhurst, line 113: nit: Rather than saying to send the GHRepository to Richie, tell the reader what piece of software you plan on sending the GHRepository to. If someone starts working on this project after you leave, they'll have no idea what this comment means.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Sami-2000: Thank you! I have actually updates the urlPattern to catch more erroneous urls, as Richie suggested, so I believe that the variable holds meaning now.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Sami-2000: Good idea! Thank you. I guess using a regular expression was unecessary here.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Sami-2000: You are right! I will ask Richie where he wants it sent.

src/main/webapp/mentor_form.html
```diff
@@ -11,14 +11,15 @@
   <h1 class="display-4 mt-5">Become a Mentor for Your Project!</h1>
   <h4 class="emphasis mb-5">Enter the URL of the GitHub repository for your project to become a mentor.</h4>
   <hr/>
-  <form>
+  <form action="javascript:submitForm()">
```
Richie78321, line 14: nit: In the spirit of [separation of concerns from the style guide](https://google.github.io/styleguide/htmlcssguide.html#Separation_of_Concerns), you could add this via a [submit event listener instead](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event): 
```javascript
form.addEventListener('submit', submitForm());
```

src/main/webapp/mentors.js
```diff
@@ -122,3 +122,32 @@ function createProjectsDiv(projectIDs) {
   }
   return projectsDiv;
 }
+
+function submitForm() {
+  const inputUrl = document.getElementById("inputRepo").value;
+  const encodedBody = new URLSearchParams();
+  encodedBody.append('inputRepo', inputUrl);
+  let url = new URL(
+      '/mentors', window.location.protocol + '//' + window.location.hostname);
```
Richie78321, line 131: nit: Use [window location origin](https://developer.mozilla.org/en-US/docs/Web/API/Location/origin) here.
```suggestion
  let url = new URL('/mentors', window.location.origin);
```

src/main/webapp/mentors.js
```diff
@@ -122,3 +122,32 @@ function createProjectsDiv(projectIDs) {
...
+    }
+    else {
+      const errorContainer = document.getElementById('error-message-container');
+      errorContainer.innerHTML = response;
+    }
+  }).catch((errorResponse) => {
+    console.error(
+        `Error ${errorResponse.statusCode}: ${errorResponse.error}`);
+      alert(errorResponse.userMessage);
+  });
```
Richie78321, line 152: `errorResponse.userMessage` is only defined for an error when using the fetch error handler. I suspect the intention was to use the fetch error handler. If so you need to wrap the fetch request:
```javascript
standardizeFetchErrors(fetchRequest, 'Fetch failed message here', 'Server error message here').then((response) => [...]);

File: src/main/webapp/mentors.js [hunk 10]
```diff
@@ -122,3 +122,32 @@ function createProjectsDiv(projectIDs) {
...
+  const fetchRequest = fetch((url), {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+    body: encodedBody,
+  });
+  
+  fetchRequest.then((response) => response.json()).then((response) => {
+    if (response == 'success') {
```
Richie78321, line 141: This definitely works, although it makes me feel like you're sort of reinventing the function of HTTP status codes. For example, if there is some sort of validation error in a user's input, you would want to respond with a 400 error. Then you can use `if (response.ok)` in a very similar way to how you're using `if (response == 'success')` here.

Richie78321: I understand that this is a product of a hybridized approach to error handling, but I'm wondering why you wouldn't want to just use the JSON-based errors for everything? Here's what it could look like if you did:
```javascript
fetchRequest.then((response) => response.json()).then((response) => {
    window.location.href = '/dashboard.html';
}).catch((errorResponse) => {
   console.error(errorResponse);
   const errorContainer = document.getElementById('error-message-container');
   errorContainer.innerText = errorResponse.userMessage;
});
```
Where all errors would be handled by the same catch block instead of in two different places. All of the other functionality stays the same.

src/main/webapp/mentors.js
```diff
@@ -122,3 +122,32 @@ function createProjectsDiv(projectIDs) {
...
+      window.location.href = '/dashboard.html';
+    }
+    else {
+      const errorContainer = document.getElementById('error-message-container');
+      errorContainer.innerHTML = response;
```
Richie78321, line 146: nit: It seems like you're only supplying text to this element. If so, it's more secure to use:
```suggestion
      errorContainer.innerText = response;
```
https://medium.com/better-programming/whats-best-innertext-vs-innerhtml-vs-textcontent-903ebc43a3fc

src/main/webapp/mentors.js
Sami-2000: I am def going to use it! I changed my js file into a module, added import statements, etc. in another pull request tho, so I will not wrap this function until that PR is merged (so I don't do the exact same thing in two PRs)

Sami-2000: Update: I did it anyways, I will just fix the merge conflict later. :/

src/main/webapp/mentors.js
Sami-2000: I did something kind of similar! Still used the catch block, but have a seperate section for improper input handling (bc I don't want it to trigger an error on the console or an alert)

PR #69, opened 2020-07-10
Add frontend signup functionality

Richie78321:
Adds the ability for a user to authenticate and create profile on the frontend.

The signup process can be broken into two distinct stages. **The important thing to consider is that the user is able to browse the website in either of these stages**, meaning that it is important to distinguish between them when writing access and authentication logic. This isn't ideal, but it is necessary after we had to abandon using OpenID due to API errors.
1. Users API authentication
2. Profile creation

Changes:
* Add `hasProfile` property to `AuthServlet` to differentiate between stages 1 and 2 in authentication.
* Update the `AuthButton` to support the stage 2 of authentication.
* Change `style.css` to allow for button colors other than purple. **The original `btn` style is now included in `btn-purple` and should be included on top of the `btn` class when making purple buttons**.
* Adds `signup.js` which drives the communication with OAuth and the Users servlet, and keeps the UI updated accordingly.

Big TODOs moving forward:
* Need to add a helper function to the users servlet that can determine if a PersonObject exists for a given User ID. This will be used for the `hasProfile` property of the `AuthServlet` response.
* Need to update the `doPost` of the Users servlet to accept the parameters sent from signup and also verify the GitHub access token.

### DIFF
--- a/src/main/java/com/google/opensesame/auth/AuthServlet.java
+++ b/src/main/java/com/google/opensesame/auth/AuthServlet.java
@@ -13,7 +13,6 @@ import javax.servlet.http.HttpServletResponse;
 
 @WebServlet("/auth")
 public class AuthServlet extends HttpServlet {
-
   public static final String AUTH_URL_REDIRECT = "/";
 
   public Gson gson = new Gson();
@@ -43,7 +42,15 @@ public class AuthServlet extends HttpServlet {
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     JsonObject responseObject = new JsonObject();
+    addAuthData(responseObject);
 
+    String responseJson = gson.toJson(responseObject);
+    response.setStatus(HttpServletResponse.SC_OK);
+    response.setContentType("application/json;");
+    response.getWriter().println(responseJson);
+  }
+
+  private void addAuthData(JsonObject responseObject) {
     UserService userService = UserServiceFactory.getUserService();
 
     responseObject.addProperty("loginUrl", userService.createLoginURL(AUTH_URL_REDIRECT));
@@ -59,14 +66,19 @@ public class AuthServlet extends HttpServlet {
       userData.addProperty("id", currentUser.getUserId());
 
       responseObject.add("user", userData);
+
+      // TODO : Add ability to check that a PersonObject exists in datastore with the user's ID.
+      // For now, will assume to be false. This can be implemented when datastore support is added
+      // for the PersonObject
+      // Checking the datastore for an existing PersonObject is mildly inefficient. This is a side
+      // effect of having to use the Users API on top of GitHub OAuth. Because there are two steps
+      // in signing up, this check ensures that the user has completed both authentication steps.
+      // If in the future OAuth can be used for backend authentication (through something like the
+      // Firebase Auth Admin SDK), this check can be removed (along with all Users API code).
+      responseObject.addProperty("hasProfile", false);
     } else {
       responseObject.addProperty("authorized", false);
+      responseObject.addProperty("hasProfile", false);
     }
-
-    String responseJson = gson.toJson(responseObject);
-
-    response.setStatus(HttpServletResponse.SC_OK);
-    response.setContentType("application/json;");
-    response.getWriter().println(responseJson);
   }
 }
--- a/src/main/java/com/google/opensesame/auth/AuthServlet.java
+++ b/src/main/java/com/google/opensesame/auth/AuthServlet.java
@@ -70,11 +70,9 @@ public class AuthServlet extends HttpServlet {
       // TODO : Add ability to check that a PersonObject exists in datastore with the user's ID.
       // For now, will assume to be false. This can be implemented when datastore support is added
       // for the PersonObject
-      // Checking the datastore for an existing PersonObject is mildly inefficient. This is a side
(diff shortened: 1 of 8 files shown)


src/main/java/com/google/opensesame/auth/AuthServlet.java
```diff
@@ -59,14 +66,19 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       responseObject.add("user", userData);
+
+      // TODO : Add ability to check that a PersonObject exists in datastore with the user's ID.
+      // For now, will assume to be false. This can be implemented when datastore support is added
+      // for the PersonObject
+      // Checking the datastore for an existing PersonObject is mildly inefficient. This is a side
+      // effect of having to use the Users API on top of GitHub OAuth. Because there are two steps
+      // in signing up, this check ensures that the user has completed both authentication steps.
+      // If in the future OAuth can be used for backend authentication (through something like the
+      // Firebase Auth Admin SDK), this check can be removed (along with all Users API code).
```
josephdhurst, line 77:
```suggestion
      // If in the future OAuth can be used for backend authentication (through something like the
      // Firebase Auth Admin SDK), we can remove the logic to check the datastore for an existing
      // PersonObject with the user's ID (along with all Users API code).
```
I don't understand the first 3 sentences in the second paragraph. The term `mildly inefficient` is vague. When I read it I think it could mean either that datastore is slow or checking for an existing PersonObject shouldn't be necessary. 

Also `This is a side effect of having to use the Users API on top of GitHub OAuth` confuses me because I don't know what `this` is. In general don't use the word `this` in technical documentation; explicitly state what you are talking about.

I think it would be sufficient to just leave the last 2 sentences but replacing the word `this` with more specific language.

src/main/webapp/signup.js
```diff
@@ -0,0 +1,162 @@
...
+  if (user) {
+    submitButton.disabled = false;
+    gitHubLinkButton.innerText = 'Unlink Your GitHub Account';
+    gitHubLinkButton.classList.add('btn-success');
+    gitHubLinkButton.classList.remove('btn-purple');
+  } else {
+    submitButton.disabled = true;
+    gitHubLinkButton.innerText = 'Link Your GitHub Account';
+    gitHubLinkButton.classList.add('btn-purple');
+    gitHubLinkButton.classList.remove('btn-success');
```
josephdhurst, line 60: nit: consider renaming `btn-success` to something more descriptive of its functionality like`btn-unlink` or `btn-signout`. ditto for `btn-purple`.

src/main/webapp/signup.js
```diff
@@ -0,0 +1,162 @@
...
+      console.error(
+        `Error ${errorResponse.statusCode}: ${errorResponse.error}`);
+      alert(errorResponse.userMessage);
+      gitHubLinkButton.disabled = false;
+      submitButton.disabled = false;
+    });
+  } else {
+    gitHubLinkButton.disabled = false;
+    submitButton.disabled = false;
+  }
```
josephdhurst, line 98:
```suggestion
  if (signupBody) {
    const signupRequest = createSignupRequest(signupBody);

    signupRequest.then((response) => {
      // If signup is successful, redirect.
      window.location.href = AFTER_SIGNUP_REDIRECT;
    }).catch((errorResponse) => {
      console.error(
        `Error ${errorResponse.statusCode}: ${errorResponse.error}`);
      alert(errorResponse.userMessage);
    });
  }
  gitHubLinkButton.disabled = false;
  submitButton.disabled = false;
```

src/main/webapp/signup.js
```diff
@@ -0,0 +1,162 @@
...
+  const gitHubLinkButton = signupForm.elements['github-link-button'];
+  const submitButton = signupForm.elements['submit-button'];
+  gitHubLinkButton.disabled = true;
+  submitButton.disabled = true;
+
+  const signupBody = createSignupBody();
+  console.log('Body of signup request:');
+  console.log(signupBody);
+  if (signupBody) {
+    const signupRequest = createSignupRequest(signupBody);
```
josephdhurst, line 81: I think you could probably move `createSignupBody` within `createSignupRequest`. That way you wouldn't have to have an extra check to see if signupBody is not null. You could just check whether createSignupRequest succeeded or raised an error.

src/main/java/com/google/opensesame/auth/AuthServlet.java
Richie78321: These are good points. I'll be more careful about vagueness and ambiguous terms like `this` moving forward. I like your suggested change as well, thanks!

src/main/webapp/signup.js
Richie78321: This is a solid suggestion, and it's directly in line with the [style guide](https://google.github.io/styleguide/htmlcssguide.html#ID_and_Class_Naming) too. However, `btn-success` is a class [tied to Bootstrap](https://getbootstrap.com/docs/4.0/components/buttons/#examples). We definitely could rename `btn-purple` though. Because this will affect more than my code, I'll discuss renaming this in our sync.

src/main/webapp/signup.js
Richie78321: I agree that this would be an ideal change, although because `signupRequest.then((response) => [...])` is asynchronous, it would mean that the buttons would be re-enabled immediately following the start of the request, not the end of the request after the response has been received. Looking at this again now, I can consolidate the ones in the promise slightly and will do that momentarily.

PR #71, opened 2020-07-11
Use standard fetch error handler.

Sami-2000:
Use Richie's fetch error handler. Call getMentors() from js file instead of html onload.


src/main/webapp/mentors.js
```diff
@@ -1,3 +1,6 @@
+import standardizeFetchErrors from '/js/fetch_handler.js';
+getMentors();
```
Richie78321, line 2: Moving `getMentors()` from onload to here changes the behavior of the code, so make sure it's intended. This function will now run when the script is loaded in the HTML (before the body is loaded) instead of after the body is loaded.

src/main/webapp/mentors.js
```diff
@@ -6,33 +9,27 @@ function getMentors() { // eslint-disable-line no-unused-vars
...
       mentorsContainer.appendChild(createMentorElement(mentor));
     }
   })
-  .catch((error) => {
-    console.log(error);
+  .catch((errorResponse) => {
+      alert(errorResponse.userMessage);
+      console.error(
+          `Error ${errorResponse.statusCode}: ${errorResponse.error}`);
   });
```
Richie78321, line 30: Nice! Just wanted to let you know that some of this behavior is going to change with #72

PR #73, opened 2020-07-13
Get User information from GitHub

Obinnabii:
- Get User information from GitHub API
- Move Datastore Entity functions and information from the ```PersonObject``` to the ```PersonBuilder``` 
to avoid using GitHub API calls where they are not necessary.
- Send user to the ```UserServlet``` with a POST method in ```user.js```


src/main/java/com/google/opensesame/servlets/PersonBuilder.java
```diff
@@ -43,4 +43,55 @@ public PersonBuilder projectIDs(ArrayList<String> projectIDs) {
...
     return this;
   }
+
+  /**
+   * Create a PersonObject instance with information from the current builder.
+   *
+   * @return PersonObject that corresponds with feilkds from this PersonBuilder instance
+   * @throws IOException
+   */
+  public PersonObject buildPersonObject() throws IOException {
```
Sami-2000, line 53: Now that you have simplified the constructors for mentors and people using the GitHub API, I wonder if a builder class is still necessary, or if this and the following functions can just be added to the Mentor and People classes. I initally created the builder so that we could have any combination of fields to construct mentors/people, but if we are only ever going to create these objects with a GitHub ID and list of projects/interests, I dont think it is still necessary.

src/main/java/com/google/opensesame/servlets/PersonObject.java
```diff
@@ -1,38 +1,81 @@
...
+   */
+  public PersonObject(String gitHubID, ArrayList<String> interestTags) throws IOException {
     this.gitHubID = gitHubID;
-    this.description = description;
     this.interestTags = interestTags;
+
+    // GitHub query
+    final GitHub gitHub = GitHubGetter.getGitHub();
+    final GHUser userGitAccount = gitHub.getUser(gitHubID);
+    this.bio = userGitAccount.getBio();
```
Sami-2000, line 34: Q: I definitely think it is worth looking into what happens to these functions when the GitHub User hasn't input certain fields. ie. if the user hasn't input a bio, what will this function return?

src/main/webapp/dashboard.js
```diff
@@ -251,6 +255,7 @@ function createMenteeCardImage(imgSrc) {
  * Call functions to populate page sections with data.
  */
 function setUpPage() {
+  // getUser('').then(createAboutMe);
```
josephdhurst, line 258:
```suggestion
```
I'm assuming you just forgot to delete this before you made the PR.

src/main/webapp/dashboard.js
```diff
@@ -1,34 +1,37 @@
+import {getUser} from 'user.js';
```
josephdhurst, line 1: The linter says `[no-unused-vars] 'getUser' is defined but never used.` It looks like this is used in the commented-out code below. Either uncomment the commented-out code or delete the commented-out code and this unused import.

src/main/java/com/google/opensesame/servlets/PersonObject.java
josephdhurst: That makes sense, maybe just put a TODO here for now though. It seems like checking these edge cases will take a significant amount of additional work.

src/main/java/com/google/opensesame/servlets/PersonBuilder.java
josephdhurst: Ditto on this. Builder objects add extra complexity, but they give you the flexibility to only set any subset of the possible fields in an object that you could set. If you'll always set the same fields and won't often need to leave any of those fields undefined in your code, it is simpler to just have one constructor.

Obinnabii: I can see your concerns. The Builder is redundant at the moment but by the next iteration of this PR, I plan to have optional fields in the Person Constructor. I will add a TODO to pivot away from a Builder in the event that It stays pointless by the next iteration of this code.

src/main/java/com/google/opensesame/servlets/PersonObject.java
Obinnabii: Will do!

PR #75, opened 2020-07-13
Create helper function for API errors

Richie78321:
Creates a helper function for sending API errors. It simply assembles the error parameters into a `JsonObject` for serialization and then sends the error as JSON along with the corresponding status code.


Richie78321:
> LGTM! Will this work with the POST methods?

Yup! Let me know if you run into any issues using it.

PR #77, opened 2020-07-13
Complete projects datastore integration

Richie78321:
Completes the projects datastore integration with the project preview data servlet.

* Changed from `fillPropertyContainer` to `fillEntity` because `Entity` class extends the functionality of `PropertyContainer` to include support for lists. Before making this fix all lists were stored as null.
* Added the ability to get a `ProjectEntity` by query.
* Added check for null lists. Datastore stores empty lists as null, so this must be checked for when the `ProjectEntity` is being constructed.
* Added a direct constructor for `ProjectPreviewData` from `ProjectEntity`. Previously this was accomplished by passing around a `Map<String, Object>` containing key-value pairs, but now that `ProjectEntity` exists as a layer of abstraction between `ProjectPreviewData` and the Datastore `Entity` class, this could be removed.
* Deleted the tests for `ProjectPreviewData` because they were testing for getter-setter functionality and required a lot of upkeep.

*Also, while debugging I came across [objectify](https://github.com/objectify/objectify), which seems to be a library that accomplishes a lot of what we are currently implementing ourselves. We will be discussing this in a meeting and might be transitioning to this in the alpha.*


src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -44,4 +55,55 @@ public static ProjectEntity getByRepositoryId(String repositoryId) {
...
+   */
+  public static ProjectEntity getByRepositoryIdOrNew(String repositoryId) {
+    ProjectEntity projectEntity = getByRepositoryId(repositoryId);
+    if (projectEntity == null) {
+      projectEntity = 
+          new ProjectEntity(repositoryId, new ArrayList<String>(), new ArrayList<String>());
+    }
+
+    return projectEntity;
+  }
```
Richie78321, line 74: Will satisfy TODO from #67

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -19,8 +24,17 @@
...
    */
   public static Key addProjectEntity(ProjectEntity projectEntity) {
+    if (projectEntity.getNumMentors() <= 0) {
+      System.out.println(
+          "Saving project ("
+              + projectEntity.repositoryId
+              + ") with zero mentors."
+              + " This is most likely a mistake, as projects should only exist in the"
+              + " datastore with one or more mentors.");
+    }
```
Obinnabii, line 34: Is this intended to go to master?

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -73,6 +73,15 @@ public ProjectEntity(
     this.repositoryId = repositoryId;
     this.mentorIds = mentorIds;
     this.interestedUserIds = interestedUserIds;
+
+    // Empty lists are represented as null in datastore, so create empty lists if null:
+    // https://cloud.google.com/appengine/docs/standard/java/datastore/entity-property-reference#Using_an_empty_list
```
Obinnabii, line 78: Interesting, I will add this to the mentors page.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
Richie78321: Yes, unless you think it's not worth including. We designed our backend with the assumption that all projects would have at least one mentor, so I wanted to include a soft check for that assumption here.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: Yeah this took quite a bit of time to debug haha. Happy to help.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -44,4 +58,56 @@ public static ProjectEntity getByRepositoryId(String repositoryId) {
...
+   */
+  public static ProjectEntity getByRepositoryIdOrNew(String repositoryId) {
+    ProjectEntity projectEntity = getByRepositoryId(repositoryId);
+    if (projectEntity == null) {
+      projectEntity =
+          new ProjectEntity(repositoryId, new ArrayList<String>(), new ArrayList<String>());
+    }
+
+    return projectEntity;
+  }
```
josephdhurst, line 78: With this new function do we still need `getByRepositoryId`? It seems like it would be simpler to just return a new ProjectEntity in getByRepositoryId when it raises a`EntityNotFoundException`. However, maybe there is some benefit to returning Null when a project isn't found that I'm not thinking about.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -44,4 +58,56 @@ public static ProjectEntity getByRepositoryId(String repositoryId) {
...
+    return projectEntities;
+  }
+
+  /**
+   * Get a list of project entities by query.
+   *
+   * @param projectsQuery The datastore query.
+   * @return Returns a list of project entities from the query.
+   */
+  public static List<ProjectEntity> getByQuery(Query projectsQuery) {
```
josephdhurst, line 110: nit: `getProjectsByQuery` would be more descriptive.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -44,4 +58,56 @@ public static ProjectEntity getByRepositoryId(String repositoryId) {
...
+  }
+
+  /**
+   * Get a list of project entities by query and fetch options.
+   *
+   * @param projectsQuery The datastore query.
+   * @param fetchOptions The fetch options for the query.
+   * @return Returns a list of project entities from the query.
+   */
+  public static List<ProjectEntity> getByQuery(Query projectsQuery, FetchOptions fetchOptions) {
```
josephdhurst, line 87: nit: getProjectsByQuery would be more descriptive.

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
Richie78321: I agree. Changing now!

src/main/java/com/google/opensesame/projects/ProjectDatastore.java
Richie78321: It's still sometimes going to be useful to know when a certain `ProjectEntity` was not found in the datastore. A good example would be when a user is making a get request to the `/project` servlet, where they must specify a specific `projectId`. In that case, if the `ProjectEntity` returned is null, the servlet knows to return an error response indicating to the user that the specified project does not exist.

PR #79, opened 2020-07-14
Integrate changes to mentor search and mentor form pages in one js file.

Sami-2000:
Add a conditional statement based on which page I am on to avoid errors.

### DIFF
--- a/src/main/webapp/mentors.js
+++ b/src/main/webapp/mentors.js
@@ -15,8 +15,7 @@ if (window.location.href.includes('mentor_form.html')) {
  */
 function getMentors() { // eslint-disable-line no-unused-vars
   console.log('entering get mentors function/n');
-  const url = new URL(
-      '/mentors', window.location.protocol + '//' + window.location.hostname);
+  const url = new URL('/mentors', window.location.origin);
   const fetchRequest = standardizeFetchErrors(
         fetch(url),
         'Failed to communicate with the server, please try again later.',
--- a/src/main/webapp/mentors.js
+++ b/src/main/webapp/mentors.js
@@ -1,4 +1,5 @@
 import standardizeFetchErrors from '/js/fetch_handler.js';
+getMentors();
 let form = document.getElementById('mentor-form');
 form.addEventListener('submit', submitForm);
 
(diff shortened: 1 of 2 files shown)


src/main/webapp/mentors.js
```diff
@@ -1,21 +1,26 @@
...
-let form = document.getElementById('mentor-form');
-form.addEventListener('submit', submitForm);
+
+if (window.location.href.includes('mentors.html')) {
+  getMentors();
+}
+if (window.location.href.includes('mentor_form.html')) {
+  console.log('mentor form page!');
+  document.getElementById('mentor-form').addEventListener('submit', submitForm);
+}
```
Obinnabii, line 9: This is a really interesting piece of code! I'm going to add some functionality to the dashboard with this!

src/main/webapp/mentors.js
```diff
@@ -1,21 +1,26 @@
...
 
 /**
  * Get mentors from the mentor servlet.
  * @param {String} response
  */
 function getMentors() { // eslint-disable-line no-unused-vars
   console.log('entering get mentors function/n');
-  const url = new URL(
-      '/mentors', window.location.protocol + '//' + window.location.hostname);
+  const url = new URL('/mentors', window.location.origin);
```
billyeh, line 18: nice

src/main/webapp/mentors.js
billyeh: Yeah, it's interesting. I wonder if it makes sense to just call both functions, and have the functions determine whether they should even do anything. It seems possible since they both rely on the presence of an element with a certain ID, right? 

So in the beginning of both functions, we could just check if that element exists, and return if not.

billyeh: Would slightly prefer that since otherwise this is a little brittle; it'll break if we move the pages to different files.

Richie78321: Yeah I agree with Bill here. Or we could move general mentor JS to a common file like `mentors.js`, and then move page-specific stuff to individual files that are only loaded on their respective pages. That way we don't have to worry about differentiating between pages but still can reuse the mentor code.

Sami-2000: Sounds good! I will check for existence instead of checking which page I am on.

PR #80, opened 2020-07-14
Add dashboard to navbar.

Sami-2000:


billyeh:
> LGTM! Maybe we can work on fixing some of the Linter issues that made it through?

Yes please! :) 

LGTM to me too otherwise.

PR #85, opened 2020-07-14
Add expanded project data support to backend

Richie78321:
Adds a class and servlet for getting full project data by project ID. 

This is being implemented to serve more detailed information about a project to the frontend. This will be used for the expanded project view, which a user will navigate to by clicking on a project preview.


src/main/java/com/google/opensesame/projects/ProjectDatastore.java
```diff
@@ -19,8 +24,17 @@
    * @return Returns the key associated with the datastore entity.
    */
   public static Key addProjectEntity(ProjectEntity projectEntity) {
+    if (projectEntity.getNumMentors() <= 0) {
+      System.out.println(
```
Obinnabii, line 28: Is this intended to get to production? If not, please add a TODO to remember to remove it

Richie78321: Good point. I'll be making the switch to objectify momentarily so it'll be gone then!

PR #86, opened 2020-07-14
Add Error Handling to the Signup and User Dashboard pipeline

Obinnabii:
- Add ```UserServlet``` Error handling to match the ```OpenSesame``` Error Handling conventions. We are transitioning to using the ```ErrorResponse``` module to send errors from servlet requests.
- Use the ```standardizedFetchErrorHandler``` function for for catching and dealing with errors recieved from the ```UserServlet``` in the ```user.js``` module.

### DIFF
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -3,6 +3,7 @@ package com.google.opensesame.servlets;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.KeyFactory;
 import com.google.appengine.api.datastore.PreparedQuery;
@@ -40,10 +41,10 @@ public class UserServlet extends HttpServlet {
     Entity userEntity; // TODO: Abstract doGet() into it's own function
     try {
       userEntity = datastore.get(userKey);
-    } catch (Exception e) {
+    } catch (EntityNotFoundException e) {
       ErrorResponse.sendJsonError(
           response,
-          e.getMessage() + "/n/n" + e.getStackTrace(),
+          "User not fount in the Datastore",
           HttpServletResponse.SC_BAD_REQUEST,
           "The user requested could not be found on the server."
               + "Please ensure that the user has an account with us.");
@@ -72,7 +73,6 @@ public class UserServlet extends HttpServlet {
   @Override
   // Send a user to datastore. Update the current information about the user if one exists.
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    try {
       String userGitHubID = request.getParameter("githubID");
       if (userGitHubID == null || userGitHubID.trim().length() == 0) {
         ErrorResponse.sendJsonError(
@@ -82,18 +82,13 @@ public class UserServlet extends HttpServlet {
             "There is no GitHub ID affiliated with this request");
         return;
       }
-      ArrayList<String> userTags =
-          new ArrayList<>(Arrays.asList(request.getParameterValues("tags")));
+      String[] userTags = request.getParameterValues("tags");
+      ArrayList<String> userTagsList = userTags == null ? new ArrayList<String>() :
+          new ArrayList<>(Arrays.asList( userTags ));
 
       Entity personEntity =
-          new PersonBuilder().gitHubID(userGitHubID).interestTags(userTags).buildPersonEntity();
+          new PersonBuilder().gitHubID(userGitHubID).interestTags(userTagsList).buildPersonEntity();
       datastore.put(personEntity);
-    } catch (Exception e) {
-      ErrorResponse.sendJsonError(
-          response,
-          e.getMessage() + "/n/n" + e.getStackTrace(),
-          HttpServletResponse.SC_BAD_REQUEST,
-          "User informatiuon could not be sent to datastore");
     }
   }
 
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -11,6 +11,8 @@ import com.google.appengine.api.datastore.Query.Filter;
 import com.google.appengine.api.datastore.Query.FilterOperator;
 import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.gson.Gson;
+import com.google.opensesame.util.ErrorResponse;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -31,19 +33,31 @@ public class UserServlet extends HttpServlet {
   }
 
   @Override
-  // Get a specific user. Return null if not found. TODO: User Validation
+  // Get a specific user. Return null if not found.
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String userGithub = request.getParameter("githubID");
     Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userGithub);
-    System.out.println(userKey);
-
+    
+    Entity userEntity;
+    try{
+    userEntity = datastore.get(userKey);
+    } catch (Exception e) {
+      ErrorResponse.sendJsonError(response, 
+      e.getMessage() + "/n/n" + e.getStackTrace(),
+      HttpServletResponse.SC_BAD_REQUEST,
+      "The user requested could not be found on the server." +
+      "Please ensure that the user has an account with us.");
+      return;
+    }    
     PersonObject userObject;
     try {
-      userObject = new PersonBuilder().buildPersonObject(datastore.get(userKey));
-      System.out.println(userObject);
+      userObject = new PersonBuilder().buildPersonObject(userEntity);
     } catch (Exception e) {
-      e.printStackTrace();
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, e.getMessage());
+      ErrorResponse.sendJsonError(response, 
+      e.getMessage() + "/n/n" + e.getStackTrace() + 
+      "/n/n PersonObject could not be instantiated from Person Entity",
+      HttpServletResponse.SC_BAD_REQUEST,
+      "User could not be instatiated in the Server");
       return;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/webapp/user.js
```diff
@@ -6,38 +11,40 @@
...
+
+  return errorFormattedFetchRequest.then((response) => response.json())
+    .then((user) => {
+      return user;
+    })
 }
 
 /**
  * Posts a user to the servlet
  * @param {URLSearchParams} userParams parameters with the users information
```
Obinnabii, line 30:
```suggestion
 * @param {URLSearchParams} userParams parameters with the users information
 * @return {Promise} response from '/user' POST
```

src/main/webapp/signup.js
```diff
@@ -4,10 +4,7 @@
  * @property {string[]} interestTags
  */
 import {gitHubAuthorizer} from './authorization.js';
-import {
-  standardizeFetchErrors,
-  makeRelativeUrlAbsolute,
-} from './js/fetch_handler.js';
+import {postUser} from '.user.js';
```
Obinnabii, line 7:
```suggestion
import {postUser} from './user.js';
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +72,29 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String userGitHubID = request.getParameter("githubID");
-    if (userGitHubID.trim().isEmpty()) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid ID");
-      return;
-    }
-    ArrayList<String> userTags = new ArrayList<>(Arrays.asList(request.getParameterValues("tags")));
+    try {
+      String userGitHubID = request.getParameter("githubID");
+      if (userGitHubID.trim().isEmpty()) {
```
Richie78321, line 77: `isEmpty()` is Java 11 only. Consider:
```suggestion
      String userGitHubID = request.getParameter("githubID").trim();
      if (userGitHubID == null || userGitHubID.length() == 0) {
```

https://stackoverflow.com/questions/32567208/best-way-to-verify-string-is-empty-or-null

src/main/webapp/signup.js
```diff
@@ -116,21 +113,7 @@ function createSignupRequest() {
...
-  });
-
-  const errorFormattedFetchRequest = standardizeFetchErrors(
-      fetchRequest,
-      'Failed to communicate with the server. Please try again later.',
-      'An error occcured while creating your account.' +
-          ' Please try again later.');
-
-  return errorFormattedFetchRequest;
+  return postUser(encodedBody);
```
Richie78321, line 116: Is there a reason why this code was moved? Will there ever be another use beyond signup where a user will need to be posted to the `/users` servlet? If your intention here is to make the post functionality more generally accessible if we intend to allow users to edit their own profiles in the future, consider following [RESTful API standards](https://www.restapitutorial.com/lessons/httpmethods.html#:~:text=The%20primary%20or%20most%2Dcommonly,but%20are%20utilized%20less%20frequently.) and using a [`doPut()`](https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/http/HttpServlet.html#doPut(javax.servlet.http.HttpServletRequest,%20javax.servlet.http.HttpServletResponse)) and sending a PUT request in that case. That will allow us to keep the two functionalities separate and follow proper standards.

src/main/webapp/user.js
```diff
@@ -6,38 +11,41 @@
...
-      .then(errorHandling).then((response) => response.json())
+  const fetchRequest = fetch(makeRelativeUrlAbsolute('/user?githubID=' + user));
+
+  const errorFormattedFetchRequest = standardizeFetchErrors(
+      fetchRequest,
+      'Failed to communicate with the server. Please try again later.',
+      'An error occcured while retrieving this account.' +
+    ' Please try again later.');
+
+  return errorFormattedFetchRequest.then((response) => response.json())
```
Richie78321, line 22: Nice!

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +32,35 @@ public void init() {
...
+          "The user requested could not be found on the server."
+              + "Please ensure that the user has an account with us.");
+      return;
+    }
     PersonObject userObject;
     try {
-      userObject = new PersonBuilder().buildPersonObject(datastore.get(userKey));
-      System.out.println(userObject);
+      userObject = new PersonBuilder().buildPersonObject(userEntity);
     } catch (Exception e) {
```
Richie78321, line 55: This looks good, although I'm wondering if we could move this functionality to a public static function like `getPersonObjectById(id)` that we could use in other places like the `hasProfile` check in the auth servlet?

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +32,35 @@ public void init() {
   }
 
   @Override
-  // Get a specific user. Return null if not found. TODO: User Validation
+  // Get a specific user. Return null if not found.
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String userGithub = request.getParameter("githubID");
```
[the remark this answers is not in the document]
Obinnabii: will do by the next PR

src/main/webapp/signup.js
Obinnabii: Oh good catch! I'll use the ```doPut()```. My reason for moving the function was to make it more accessible but I guess it won't really be used anywhere else. That being said, I'd still like to have it in the ```user.js``` module. I like having all the functions that use the ```/user``` servlet in the same place. If you think it's better in another module, that's fine too

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +72,29 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      Entity personEntity =
+          new PersonBuilder().gitHubID(userGitHubID).interestTags(userTags).buildPersonEntity();
+      datastore.put(personEntity);
+    } catch (Exception e) {
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage() + "/n/n" + e.getStackTrace(),
+          HttpServletResponse.SC_BAD_REQUEST,
+          "User informatiuon could not be sent to datastore");
+    }
```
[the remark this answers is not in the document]
Obinnabii: Interesting point! do we want to expose really specific error messages to the user? I'll add more try blocks. As it is, when an error occurs, we get a specific error message with an exposed callstack returned to the client but the user gets a vague ```"user information could not be sent to datastore"``` alert

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: will do!

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +32,35 @@ public void init() {
...
 
   @Override
-  // Get a specific user. Return null if not found. TODO: User Validation
+  // Get a specific user. Return null if not found.
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
     String userGithub = request.getParameter("githubID");
     Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userGithub);
-    System.out.println(userKey);
 
+    Entity userEntity;
```
Obinnabii, line 40: > This looks good, although I'm wondering if we could move this functionality to a public static function like `getPersonObjectById(id)` that we could use in other places like the `hasProfile` check in the auth servlet?
```suggestion
    Entity userEntity; // TODO: Abstract doGet() into it's own function 
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +72,29 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      if (userGitHubID == null || userGitHubID.trim().length() == 0) {
+        ErrorResponse.sendJsonError(
+            response,
+            "Empty GitHub ID",
+            HttpServletResponse.SC_BAD_REQUEST,
+            "There is no GitHub ID affiliated with this request");
+        return;
+      }
+      ArrayList<String> userTags =
+          new ArrayList<>(Arrays.asList(request.getParameterValues("tags")));
```
billyeh, line 86: Does this part need to be in the `try` block? Best to keep the scope of it as limited as possible and move these lines out if not.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +32,35 @@ public void init() {
...
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String userGithub = request.getParameter("githubID");
+    String userGithub = request.getParameter("githubID"); // TODO:Check for null values
     Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userGithub);
-    System.out.println(userKey);
 
+    Entity userEntity; // TODO: Abstract doGet() into it's own function
+    try {
+      userEntity = datastore.get(userKey);
+    } catch (Exception e) {
```
billyeh, line 43: Is it possible to be a little more specific about what Exceptions we expect to be thrown here? For example, it should just be EntityNotFoundException here. https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/EntityNotFoundException.html

And below in the other try/catches.

src/main/webapp/signup.js
Richie78321: As long as we're on the same page that it really will only be used for signup, I'd prefer if it stayed in the original file `signup.js`. That way we can avoid confusion with the post request vs. put request and we can have more freedom with changes to this function down the line. Also, when you moved it, you exposed a lot of the error checks and content limiting. Before one was only able to make a post request with specific signup body data, and now a post request can be made with any URLEncoded data.

Obinnabii, line 116:
```suggestion
 const fetchRequest = fetch(makeRelativeUrlAbsolute('/user'), {
      method: 'POST',	
    headers: {	
      'Content-Type': 'application/x-www-form-urlencoded',	
    },	
    body: encodedBody,	
  });	

  const errorFormattedFetchRequest = standardizeFetchErrors(	
      fetchRequest,	
      'Failed to communicate with the server. Please try again later.',	
      'An error occcured while creating your account.' +	
          ' Please try again later.');	

  return errorFormattedFetchRequest;
```

Obinnabii:
I am going to overhaul this User Servlet implementation entirely with Objectify

PR #87, opened 2020-07-14
Add email field to user signup

Richie78321:
Add an email field to user signup. This includes basic syntactic validation of the email on the client side, but **should also be validated on the server-side before the client data is saved**. 

This field was added following a design decision to avoid using the email associated with someone's GitHub account. We hope to provide more flexibility to users by allowing them to choose what email to use as their point of contact.

### DIFF
--- a/src/main/webapp/signup.js
+++ b/src/main/webapp/signup.js
@@ -2,6 +2,7 @@
  * @typedef SignUpData
  * @property {string} gitHubAuthToken
  * @property {string[]} interestTags
+ * @property {string} emailAddress
  */
 import {gitHubAuthorizer} from './authorization.js';
 import {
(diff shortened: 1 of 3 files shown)


src/main/webapp/signup.js
```diff
@@ -158,7 +167,19 @@ function createSignupBody() {
...
+/**
+ * Validates an email syntactically against a regular expression.
+ * Sourced from:
+ * https://stackoverflow.com/questions/46155/how-to-validate-an-email-address-in-javascript
+ * @param {string} email
+ */
+function validateEmail(email) {
+  const re = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
+  return re.test(String(email).toLowerCase());
+}
```
Obinnabii, line 183: Do we need this?

Richie78321: No we don't. I did some further research and it appears that trying to validate emails with regex is generally considered bad practice. Enough validation is already happening through the browser by default by including the `type="email"` in the input tag.

PR #91, opened 2020-07-15
Improve communication between Authentication, Dashboard and user.js module

Obinnabii:
- Add an ```email``` field to the ```PersonObject```.


src/main/java/com/google/opensesame/servlets/PersonObject.java
```diff
@@ -22,17 +22,19 @@
    *
    * @param gitHubID
    * @param interestTags
+   * @param email
    * @throws IOException
    */
-  public PersonObject(String gitHubID, ArrayList<String> interestTags) throws IOException {
+  public PersonObject(String gitHubID, ArrayList<String> interestTags, String email)
+      throws IOException {
```
Richie78321, line 29: Can we also add the userId to the person object? I have a feeling a lot of this code is going to be torn down on the transition to Objectify so feel free to add it then (I think you need to add the ID for Objectify, see [here](https://github.com/objectify/objectify/wiki/Entities#the-basics)):

> Entities must have one field annotated with @Id. The actual name of the field is irrelevant and can be renamed at any time, even after data is persisted. This value (along with the kind 'Car') becomes part of the Key which identifies an entity.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +35,51 @@ public void init() {
...
+  // supplied.
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String userGithub = request.getParameter("githubID");
-    Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userGithub);
-    System.out.println(userKey);
 
+    String userID =
+        request.getParameter("userID") != null
+            ? request.getParameter("userID")
+            : AuthServlet.getAuthorizedUser().getUserId();
```
Richie78321, line 45: Could we make this into an if statement? I think it would be much more readable:
```suggestion
    String userID = request.getParameter("userID");
    if (userID == null) {
      userID = AuthServlet.getAuthorizedUser().getUserId();
    }
```

Also keep in mind that in your current implementation and in this suggested implementation you are at risk of a null pointer exception here if no `userID` is provided with the request and the user is not authenticated, in which case `getAuthorizedUser()` will return null.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +35,51 @@ public void init() {
...
+    try {
+      userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userID);
+    } catch (Exception e) {
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage() + "/n/n" + e.getStackTrace(),
+          HttpServletResponse.SC_BAD_REQUEST,
+          "An error occured while getting the User's profile");
+      return;
+    }
```
Richie78321, line 57: Is the HTTP 400 error code the most appropriate error code here? Looking at the user message here, I think a 500 error (or some server error) is more appropriate. A 400 error implies that the error can be fixed by actions from the user.

Richie78321: Also I'm not sure if I understand why there is a try-catch here. It doesn't seem like [createKey](https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/KeyFactory.html#createKey-java.lang.String-java.lang.String-) throws any meaningful errors.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,19 +35,51 @@ public void init() {
...
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage()
+              + "/n/n"
+              + e.getStackTrace()
+              + "/n/n PersonObject could not be instantiated from Person Entity",
+          HttpServletResponse.SC_BAD_REQUEST,
+          "User could not be instatiated in the Server");
       return;
     }
```
Richie78321, line 84: See comment above about appropriate HTTP status codes. I'm thinking that this is another case where a 5xx error would be most appropriate.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +91,57 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      userGitHubAuthToken = request.getParameter("gitHubAuthToken");
+      userTags = new ArrayList<>(Arrays.asList(request.getParameterValues("interestTags")));
+    } catch (Exception e) {
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage() + "/n/n" + e.getStackTrace() + "/n/n Incomplete POST body parameters",
+          HttpServletResponse.SC_BAD_REQUEST,
+          "User information was incomplete");
+      return;
+    }
```
Richie78321, line 107: Why is this in a try-catch block? [No error is ever thrown](https://tomcat.apache.org/tomcat-5.5-doc/servletapi/javax/servlet/ServletRequest.html#getParameter(java.lang.String)) from `request.getParameter("gitHubAuthToken");` and we could avoid an error being thrown in `userTags = new ArrayList<>(Arrays.asList(request.getParameterValues("interestTags")));` by checking first if the parameter is null and sending an error response then.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +91,57 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    // Get User information from GitHub using the Oath token.
+    GHMyself userGHMyself;
+    try {
+      GitHub userPersonalGitHubInstance = GitHub.connectUsingOAuth(userGitHubAuthToken);
+      userGHMyself = userPersonalGitHubInstance.getMyself();
+    } catch (Exception e) {
```
Richie78321, line 113: Nice

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +91,57 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage()
+              + "/n/n"
+              + e.getStackTrace()
+              + "/n/n User informatiuon could not be sent to datastore",
+          HttpServletResponse.SC_BAD_REQUEST,
+          "User information could not be sent to datastore");
       return;
     }
```
Richie78321, line 144: I think this is another instance where the error being returned to the user is "unactionable," and I think it also merits an error code in the 5xx region instead of the error code for bad request. When do we anticipate encountering an error here, and can we more specifically handle that error with a response?

src/main/webapp/dashboard.js
```diff
@@ -236,7 +235,7 @@ function createMenteeCardImage(imgSrc) {
  * Call functions to populate page sections with data.
  */
 function setUpPage() {
-  getUser(dummyGitHubID).then(createAboutMe);
+  getUser().then(createAboutMe);
```
Richie78321, line 238: You need a catch block here in case the fetch request returns any error. Consider how you would like errors to be displayed to the user.

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: It throws an error If the UserID is null. The Exception is not in the documentation for some reason.

src/main/java/com/google/opensesame/servlets/PersonObject.java
josephdhurst: On this note, are you planning on deprecating PersonObject eventually? It seems that PersonObject and PersonEntity are nearly identical so we might only need one.

Obinnabii:
I am overhauling the entire system to run with Objectify

PR #92, opened 2020-07-15
Add function to check if a user has a profile with OpenSesame

Obinnabii:
An important consideration Is whether this function should throw an ```Exception``` on invalid inputs

### DIFF
--- a/src/main/java/com/google/opensesame/servlets/UserServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/UserServlet.java
@@ -3,6 +3,7 @@ package com.google.opensesame.servlets;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.EntityNotFoundException;
 import com.google.appengine.api.datastore.Key;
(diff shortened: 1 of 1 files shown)


src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -87,4 +87,19 @@ public PreparedQuery queryInDatabase(
   }
 
   // TODO: make error handling conform with Richie's error handling
+  /**
+   * Check if a user is stored with userID in Datastore. Invalid return false.
```
billyeh, line 91:
```suggestion
   * Checks if a user is stored with userID in Datastore. If invalid, return false.
```
nit

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -87,4 +87,19 @@ public PreparedQuery queryInDatabase(
...
+   *
+   * @param userID
+   * @return true if userID exists in datastore, false otherwise.
+   */
+  public Boolean hasProfile(String userID) {
+    try {
+      Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userID);
+      datastore.get(userKey);
+    } catch (Exception e) {
```
billyeh, line 100: Again, if we could be a bit more specific about which exceptions we're catching here, that would be great.

PR #94, opened 2020-07-15
Add Objectify dependency and create mentor and person entity classes.

Sami-2000:
Add web filter and bootstrapper necessary to configure version 5.1.1 of Objectify.


src/main/java/com/google/opensesame/objectify/ObjectifyWebFilter.java
```diff
@@ -0,0 +1,7 @@
+package com.google.opensesame.objectify;
+
+import com.googlecode.objectify.ObjectifyFilter;
+import javax.servlet.annotation.WebFilter;
+
+@WebFilter(urlPatterns = {"/*"})
```
Obinnabii, line 6: Maybe we can look into what Servlets need to be  Filtered? Not a big deal tbh, this works fine.

src/main/java/com/google/opensesame/servlets/PersonEntity.java
```diff
@@ -0,0 +1,28 @@
...
+    this.interestTags = person.getTags();
+    this.email = person.getEmail();
+  }
+
+  public PersonEntity(String id, String gitHubId, ArrayList<String> interestTags, String email) {
+    this.id = id;
+    this.gitHubId = gitHubId;
+    this.interestTags = interestTags;
+    this.email = email;
+  }
```
josephdhurst, line 27: Can we deprecate PersonObject and just use `PersonEntity` instead now? Even better, could we just modify `PersonObject` slightly so that it works with Objectify? It seems like both the objects store the same data, so I feel like one may be enough.

src/main/java/com/google/opensesame/servlets/MentorEntity.java
```diff
@@ -0,0 +1,21 @@
...
+
+  private MentorEntity() {
+    super();
+  }
+
+  public MentorEntity(
+      PersonEntity person, ArrayList<Long> projectIDs, ArrayList<String> menteeIDs) {
+    super(person.id, person.gitHubId, person.interestTags, person.email);
+    this.projectIDs = projectIDs;
+    this.menteeIDs = menteeIDs;
```
josephdhurst, line 19: The comment I made about deprecating `PersonObject` also applies to `MentorObject`.

src/main/java/com/google/opensesame/servlets/PersonEntity.java
Sami-2000: Hi Joe! So one of the main reasons we still have Objects and Entities is due to the GitHub API call limit. We don't want to store all of the parameters we get from GitHub (eg. name, location, bio, profile pic) in datastore, so we want to keep the Entities as light as possible. However, we also don't want to have to query GitHub every time we want one of these parameters. Obi and I thought that, by turning a light Entity into a Robust object once using the GitHub API, we will be able to have direct access to all parameters without having to make repeated API calls.

src/main/java/com/google/opensesame/objectify/ObjectifyWebFilter.java
Richie78321: I'm thinking we should keep this as-is. Otherwise it'll require constant upkeep as new servlets are introduced

PR #96, opened 2020-07-15
Patch UserServlet Compilation Error

Obinnabii:
- Patch an error that made its way into master


Obinnabii:
Thanks guys
I almost had a heart attack with that one.

PR #97, opened 2020-07-15
Objectify for mentors

Sami-2000:
Use objectify in the mentors servlet. Some areas still commented out for now as I await merges by Obi and RIchie.


src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -90,28 +66,51 @@ public void doPost(HttpServletRequest request, HttpServletResponse response)
     String repoUrl = request.getParameter("inputRepo");
     String repoName = repoUrl.replaceFirst("https://github.com/", "");
```
Obinnabii, line 67: We might want to use a URL parser here. The ```repoURL``` might be in the form "https://github.com/" it could be "github.com/" or some other valid iteration of a URL. I think it would make sense to use a parser that is designed for URL parsing to avoid unwanted rigidity. Not urgent but something we should look into. [URL](https://docs.oracle.com/javase/tutorial/networking/urls/urlInfo.html)  is a well-maintained Java Class that we can use. We can get the host and the path easily.

PR #98, opened 2020-07-16
Add React routing for project frontend.

Richie78321:
Add [React router](https://reactrouter.com/) on the projects frontend to allow for navigation to an expanded project view. This means that users can now click on a project preview card and be seamlessly redirected to an expanded project view, which can be bookmarked and linked to like a real URL.

* Add a generalized fetching component called `DataFetcher` to avoid code repeat. This will be added to existing components that use fetch requests in the future to consolidate code.
* Add a basic `ExpandedProject` frontend component that will be further implemented in a future PR.
* Updates the `ProjectSearch` component for React routing and moves the existing project preview loading to the `DataFetcher` component.


src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Called after the component is first rendered by React.
+   * @override
+   */
+  componentDidMount() {
+    // TODO : Add automatic retry functionality here.
+    // TOOD : Add better user notification system here.
+    this.props.createFetchRequest(this.abortController.signal).then((data) => {
+      console.log('Data received:');
+      console.log(data);
```
Obinnabii, line 48: Will this go into production?

src/main/webapp/js/projects/react/components/ExpandedProject.js
```diff
@@ -0,0 +1,28 @@
+import {expandedProjectType} from '../prop_types.js';
+
+/**
+ * An expanded view of a project.
+ * @param {Object} props
+ * @return {React.Component} Returns the React component.
+ */
+export default function ExpandedProject(props) {
```
Obinnabii, line 8: is it neccesary to use default exports?

src/main/webapp/js/projects/react/components/ExpandedProject.js
```diff
@@ -0,0 +1,28 @@
...
+  if (props.loading) {
```
Obinnabii, line 9: Very interesting stuff. When does this check for the loading property get refreshed?

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,71 @@
+import checkTesting from '../../../checkTesting.js';
+import {basicErrorHandling} from '../../../fetch_handler.js';
+checkTesting();
+
+/**
+ * A generic data fetching component that can be used for fetching data for any
+ * component.
+ */
+export default class DataFetcher extends React.Component {
```
billyeh, line 9: Also recommending we follow style guide and avoid default export here.

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,71 @@
...
+  componentDidMount() {
+    // TODO : Add automatic retry functionality here.
+    // TOOD : Add better user notification system here.
+    this.props.createFetchRequest(this.abortController.signal).then((data) => {
+      console.log('Data received:');
+      console.log(data);
+
+      this.setState({
+        isFetching: false,
+        data,
```
billyeh, line 52: Learned that this works now in javascript. Nice

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,71 @@
+import checkTesting from '../../../checkTesting.js';
+import {basicErrorHandling} from '../../../fetch_handler.js';
+checkTesting();
+
+/**
+ * A generic data fetching component that can be used for fetching data for any
+ * component.
```
billyeh, line 7: For such a generic component, I think it doesn't hurt to point to or provide a usage example in the comments.

billyeh: Need to describe "What does this do for me" and "How do I use it"

Richie78321: That's a good point, as I definitely intend to this to be used by more people than just me. I'll improve the documentation for this.

src/main/webapp/js/common/react/components/DataFetcher.js
Richie78321: Okay! [That part](https://google.github.io/styleguide/jsguide.html#es-module-exports) of the style guide definitely slipped by me, so I'll need to update all of the React components to follow the style guide. I'll do that in another PR, but for now I'll change the DataFetcher from `export default` in this PR.

src/main/webapp/js/projects/react/components/ExpandedProject.js
Richie78321: Good question! When the `DataFetcher` component finishes fetching the data, it will update its state using `this.setState([...])`. When the state is updated this way, it [triggers a re-render](https://reactjs.org/docs/react-component.html#setstate) of the component:

> setState() enqueues changes to the component state and tells React that this component and its children need to be re-rendered with the updated state.

Once the `DataFetcher` component is re-rendered, it will also re-render the `ExpandedProject` component, which will then re-run the function that you see above.

src/main/webapp/js/projects/react/components/ExpandedProject.js
Richie78321: It's not necessary. It's just React convention, although I'm realizing from Bill's message below that `export default` is not good style according to the [style guide](https://google.github.io/styleguide/jsguide.html#es-module-exports) so I'll be removing this.

src/main/webapp/js/common/react/components/DataFetcher.js
Richie78321: No, I'll add a TODO to remove it before production. Thanks!

src/main/webapp/js/projects/react/components/ExpandedProject.js
Obinnabii: I see, the whole cascading thing. very cool!

src/main/webapp/js/common/react/components/DataFetcher.js
billyeh: Yep! Sometimes, the more general your code, the higher the comment/code ratio :) I think I remember we looked at some examples in the Google codebase where there's about 3 lines of code and 100s of comments to explain all the usage, implications, etc.

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,122 @@
...
+ * @property {boolean} isFetching Whether or not the data is currently being
+ *    fetched.
+ * @property {?Object} data The data received from the fetch request. This is
+ *    null if the data is still being fetched.
+ */
+/**
+ * @typedef DataFetcherProps
+ * @property {function(AbortSignal): Promise} createFetchRequest
+ * @property {function(DataFetcherState): React.Component} render
+ */
```
Richie78321: @billyeh How does this look for documentation?

src/main/webapp/js/common/react/components/DataFetcher.js
Richie78321: That makes sense -- the more general the code the more people are likely to use it. And yeah I definitely remember looking at code like that haha.

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,121 @@
...
+ * 1) A function that returns a fetch request for the data needed. This will be
+ * provided to the DataFetcher through its props. See {@link DataFetcherProps}.
+ * One important thing to note is that the DataFetcher supports abort signalling,
+ * which is used to cancel the fetch request in the case that the component is
+ * removed before the fetch can be completed. See this for how to use the abort
+ * signal in your fetch request:
+ * https://stackoverflow.com/questions/31061838/how-do-i-cancel-an-http-fetch-request
+ * 2) A function that renders the React component using the data being fetched.
+ * This follows the render prop design pattern:
+ * https://reactjs.org/docs/render-props.html
```
billyeh, line 16:
```suggestion
```
Think this is fine if we have the same documentation further down in the `render` function.

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,121 @@
...
+ * boilerplate.
+ *
+ * To use this component, you need two things:
+ * 1) A function that returns a fetch request for the data needed. This will be
+ * provided to the DataFetcher through its props. See {@link DataFetcherProps}.
+ * One important thing to note is that the DataFetcher supports abort signalling,
+ * which is used to cancel the fetch request in the case that the component is
+ * removed before the fetch can be completed. See this for how to use the abort
+ * signal in your fetch request:
+ * https://stackoverflow.com/questions/31061838/how-do-i-cancel-an-http-fetch-request
```
billyeh, line 13:
```suggestion
 * removed before the fetch can be completed.
```
Info is already present below.

src/main/webapp/js/common/react/components/DataFetcher.js
```diff
@@ -0,0 +1,121 @@
+/**
+ * @fileoverview This component is used to fetch data for another component.
+ * Using this component means that you do not need to create your own fetch
+ * boilerplate.
+ *
+ * To use this component, you need two things:
+ * 1) A function that returns a fetch request for the data needed. This will be
+ * provided to the DataFetcher through its props. See {@link DataFetcherProps}.
+ * One important thing to note is that the DataFetcher supports abort signalling,
```
billyeh, line 9: Let's fix the line lengths to keep the linter happy :)

src/main/webapp/js/projects/react/components/ProjectPreview.js
```diff
@@ -28,7 +34,7 @@ export default function ProjectPreview(props) {
             <strong className="h5 text-success mr-2">
               {projectPreview.numMentors}
             </strong>
-            Mentors
+              Mentors
```
billyeh, line 37:
```suggestion
            Mentors
```
Do we need to indent this more? It's still within the `span`, right?

src/main/webapp/js/projects/react/components/ProjectsSearch.js
```diff
@@ -27,38 +34,109 @@ export default class ProjectsSearch extends React.Component {
...
   }
 }
+
+/**
+ * A data fetcher for project preview data that renders the list of project
+ * previews after the data is loaded.
+ * @return {React.Component} Returns the React component.
+ */
+function ProjectPreviewFetcher() {
+  const onRender = (dataFetcher) => {
```
billyeh, line 64:
```suggestion
  const onRender = (props) => {
```
Should this be `props`? It was really confusing to me, I know technically it doesn't matter what we call it but it's quite a deparature from React conventions if they are props that we don't call props :)

src/main/webapp/js/projects/react/components/ProjectsSearch.js
```diff
@@ -27,38 +34,109 @@ export default class ProjectsSearch extends React.Component {
...
+}
+
+/**
+ * A data fetcher for project expanded project data that renders the expanded
+ * project view after the data is loaded.
+ * @return {React.Component} Returns the React component.
+ */
+function ExpandedProjectFetcher() {
+  const {projectId} = ReactRouterDOM.useParams();
+  const onRender = (dataFetcher) => {
```
billyeh, line 86:
```suggestion
  const onRender = (dataFetcherState) => {
```

src/main/webapp/js/projects/react/components/ProjectsSearch.js
```diff
@@ -6,8 +6,15 @@ import ProjectList from './ProjectList.js';
 import {
   standardizeFetchErrors,
   makeRelativeUrlAbsolute,
-  basicErrorHandling,
 } from '../../../fetch_handler.js';
+import {DataFetcher} from '../../../common/react/components/DataFetcher.js';
+import ExpandedProject from './ExpandedProject.js';
+
+// Using HashRouter to avoid having to configure servlet routing:
+// https://stackoverflow.com/questions/27928372/react-router-urls-dont-work-when-refreshing-or-writing-manually
```
billyeh, line 14: Comments including links like this are helpful, though note previous guidance about comments explaining why/what the code is trying to accomplish, rather than trying to explain what the code is doing.

https://screenshot.googleplex.com/v62RVOLU2KZ

Not sure if this would fall into the category of "assume readers are familiar with React Router" which actually most of us working on this project aren't. But I'd consider putting helpful links like this in the PR rather than in the comments depending on how important you think it is for it to be right there with the code.

billyeh:
Really cool that you seem to always come up with reusable libraries/functions to reduce boilerplate!

Just a little note/FYI: the reusable libraries are great, but a big part of their success is not just how well they're written or how useful they are to you, etc. You need to convince people "this piece of code is useful enough for me to want to learn how to use and it integrate it into mine." Writing documentation comments, showing examples how to use it, etc. are all a big part of that.

You also need to be prepared for people to say, "No, I don't want to use that library/function." Just because you wrote it doesn't mean everyone automatically must migrate to use it. Perfectly valid reasons to say something like that are, e.g., "that code would add a dependency that I'm not sure I can trust/I don't want that dependency", "I don't think it's worth onboarding our developers to use that code", "it doesn't look easy to use", etc. etc. 

So as the writer of the library/functions, you're free to to suggest that others use it, but it's your job to show them how and what the benefits are :D And sometimes people will say no anyway.

Richie78321:
> Really cool that you seem to always come up with reusable libraries/functions to reduce boilerplate!
> 
> Just a little note/FYI: the reusable libraries are great, but a big part of their success is not just how well they're written or how useful they are to you, etc. You need to convince people "this piece of code is useful enough for me to want to learn how to use and it integrate it into mine." Writing documentation comments, showing examples how to use it, etc. are all a big part of that.
> 
> You also need to be prepared for people to say, "No, I don't want to use that library/function." Just because you wrote it doesn't mean everyone automatically must migrate to use it. Perfectly valid reasons to say something like that are, e.g., "that code would add a dependency that I'm not sure I can trust/I don't want that dependency", "I don't think it's worth onboarding our developers to use that code", "it doesn't look easy to use", etc. etc.
> 
> So as the writer of the library/functions, you're free to to suggest that others use it (or, best: discuss with the people on the project, then make the changes to use the library yourself if everyone is ok with it), but it's your job to show them how and what the benefits are :D And sometimes people will say no anyway.

Yup! I agree with everything you're saying. Just to clarify, this is boilerplate meant mostly for me, and I truly don't mean to impose with it. I'm the only person really using React at the moment, so it hopefully shouldn't pressure anyone to do anything. I only commented it a lot because I could see it being useful if people needed to use React in the future.

Thanks for the tips!

billyeh:
> > Really cool that you seem to always come up with reusable libraries/functions to reduce boilerplate!
> > Just a little note/FYI: the reusable libraries are great, but a big part of their success is not just how well they're written or how useful they are to you, etc. You need to convince people "this piece of code is useful enough for me to want to learn how to use and it integrate it into mine." Writing documentation comments, showing examples how to use it, etc. are all a big part of that.
> > You also need to be prepared for people to say, "No, I don't want to use that library/function." Just because you wrote it doesn't mean everyone automatically must migrate to use it. Perfectly valid reasons to say something like that are, e.g., "that code would add a dependency that I'm not sure I can trust/I don't want that dependency", "I don't think it's worth onboarding our developers to use that code", "it doesn't look easy to use", etc. etc.
> > So as the writer of the library/functions, you're free to to suggest that others use it (or, best: discuss with the people on the project, then make the changes to use the library yourself if everyone is ok with it), but it's your job to show them how and what the benefits are :D And sometimes people will say no anyway.
> 
> Yup! I agree with everything you're saying. Just to clarify, this is boilerplate meant mostly for me, and I truly don't mean to impose with it. I'm the only person using React at the moment, so it hopefully shouldn't pressure anyone to do anything. I only commented it a lot because I could see it being useful if people needed to use React in the future.
> 
> Thanks for the tips!

Haha thanks for reading the wall of text, don't mean to imply you're imposing on anyone! Just some thoughts I had :) Great job!

Richie78321:
> Haha thanks for reading the wall of text, don't mean to imply you're imposing on anyone! Just some thoughts I had :) Great job!

Ah the ol' loss of context from written language haha. Glad to hear that wasn't by intention

src/main/webapp/js/projects/react/components/ProjectPreview.js
Richie78321: Agreed, will make the change now.

src/main/webapp/js/projects/react/components/ProjectsSearch.js
Richie78321: Yeah it's not technically props but I can see the resemblance. `dataFetcherState` definitely makes it more clear and I'll make that change! Thanks.

src/main/webapp/js/projects/react/components/ProjectsSearch.js
Richie78321: That's a good point. I originally included the link because for those familiar with React Router they would have some immediate questions as to why I'm using the HashRouter here instead of the BrowserRouter, which is preferred. I'll take a pass over the whole file with your advice in mind, thanks for the help with this!

PR #99, opened 2020-07-16
Improve communication between Authentication, Dashboard and user.js module using Objectify

Obinnabii:
- Add an ```email``` field to the ```PersonObject``` and the dashboard.
- Modify the ```PersonBuilder``` to work with ```Objectify```


src/main/java/com/google/opensesame/servlets/MentorObject.java
```diff
@@ -6,19 +6,35 @@
...
+  private ArrayList<Long> projectIDs;
+  private ArrayList<String> menteeIDs;
 
-  public MentorObject(String gitHubID, ArrayList<String> interestTags, ArrayList<String> projectIDs)
+  public MentorObject(
+      String userID,
+      String gitHubID,
+      ArrayList<String> interestTags,
+      ArrayList<String> menteeIDs,
+      ArrayList<Long> projectIDs,
```
Richie78321, line 17: No biggie, but one thing to consider here is that in my [ProjectEntity class](https://github.com/googleinterns/step22-2020/blob/master/src/main/java/com/google/opensesame/projects/ProjectEntity.java) I'm storing `projectID`s as strings instead of longs. I'm aware that they come from the GitHub Java API as longs, and I can elaborate on why I chose strings instead of longs. For compatibility's sake I'm wondering if we should discuss, pick one data type, and stick with it.

src/main/java/com/google/opensesame/servlets/PersonBuilder.java
```diff
@@ -38,34 +41,55 @@ public PersonBuilder interestTags(ArrayList<String> interestTags) {
...
    * @return PersonObject that corresponds to the entity retrieved from datastore
    * @throws IOException
    */
-  public PersonObject buildPersonObject(Entity personEntity) throws IOException {
-    String entityGitHubID = (String) personEntity.getProperty(GITHUB_ID_FIELD);
-    this.gitHubID(entityGitHubID);
-    ArrayList<String> entityTagList = (ArrayList<String>) personEntity.getProperty(TAG_LIST_FIELD);
-    this.interestTags(entityTagList);
-    return this.buildPersonObject();
+  public PersonBuilder fromEntity(PersonEntity personEntity) throws IOException {
```
Richie78321, line 82: Nice! I'm loving the naming consistency with this and the projects stuff. Will definitely help to keep things readable!

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -1,27 +1,32 @@
...
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import org.kohsuke.github.GHMyself;
+import org.kohsuke.github.GitHub;
 
 @WebServlet("/user")
 public class UserServlet extends HttpServlet {
 
+  private String pageToRedirectToIfUserNotAuthenticated = "index.html";
```
Richie78321, line 28: nit: Consider something a little shorter, this is a bit hard to read imo.

```suggestion
  private String noAuthenticationRedirect = "index.html";
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,22 +36,36 @@ public void init() {
...
+        userID =
+            AuthServlet.getAuthorizedUser().getUserId(); // TODO: Should this be a functionality?
+      } catch (NullPointerException e) {
+        ErrorResponse.sendJsonError(
+            response,
+            "UserID not Supplied and user not logged in",
+            HttpServletResponse.SC_FORBIDDEN,
+            "You are not logged in");
+        return; // TODO: Establish Redirect page path
+      }
```
Richie78321, line 55: nit: Consider using an if statement like
```java
User user = AuthServlet.getAuthorizedUser();
if (user == null) {
  [...]
}
```
instead of the try catch block. Not only do I think it's more readable that way, but it seems that the unspoken rule for try-catches is [this](https://softwareengineering.stackexchange.com/questions/141522/if-else-statements-or-exceptions):

> Typically, I tend to use exceptions for exceptional cases - the instance where nothing should go wrong, and if it does go wrong, that's a big problem.

Where a user not being authorized is a rather common occurrence.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,22 +36,36 @@ public void init() {
...
+    }
+    PersonEntity userEntity = ofy().load().type(PersonEntity.class).id(userID).now();
```
Richie78321, line 57: Wow so nice

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -31,22 +36,36 @@ public void init() {
...
+    if (userEntity == null) {
+      ErrorResponse.sendJsonError(
+          response,
+          "User retrieved from Datastore was null",
+          HttpServletResponse.SC_NOT_FOUND,
+          "User does not exist.");
```
Richie78321, line 63: Nice! Good use of `SC_NOT_FOUND`.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +74,45 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-    if (userGitHubID.trim().isEmpty()) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid ID");
-      return;
+    String userGitHubAuthToken = request.getParameter("gitHubAuthToken");
+    ArrayList<String> userTags =
+        new ArrayList<>(
+            Arrays.asList(
+                request.getParameterValues("interestTags") == null
+                    ? new String[] {}
+                    : request.getParameterValues("interestTags")));
```
Richie78321, line 83: nit: Consider breaking this up into multiple lines. At least for me, it's a bit hard to read.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +74,45 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+    try {
+      userID = AuthServlet.getAuthorizedUser().getUserId();
+    } catch (NullPointerException e) {
+      ErrorResponse.sendJsonError(
+          response,
+          "User not logged in",
+          HttpServletResponse.SC_FORBIDDEN,
+          "You are not logged in");
+      return; // TODO: Establish Redirect page path
     }
```
Richie78321, line 95: Same comment as above about using an if-statement

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +74,45 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   @Override
   // Send a user to datastore. Update the current information about the user if one exists.
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String userGitHubID = request.getParameter("githubID");
-    if (userGitHubID.trim().isEmpty()) {
-      response.sendError(HttpServletResponse.SC_BAD_REQUEST, "Invalid ID");
-      return;
+    String userGitHubAuthToken = request.getParameter("gitHubAuthToken");
```
Richie78321, line 77: nit: In the future it might be worth checking to see if this is null before using it with the GitHub interface. Omitting the auth token should most likely raise a different error than if the user is supplying an auth token that is invalid.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -55,16 +74,45 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      ErrorResponse.sendJsonError(
+          response,
+          e.getMessage()
+              + e.getStackTrace()
+              + "Could not get authenticated User with the given token",
+          HttpServletResponse.SC_UNAUTHORIZED,
+          "User could not be authenticated by GitHub, please try again");
+      return;
+    }
+    // Build and send the User's datastore entity
```
Richie78321, line 111: Does this mean to say "Build and save," or am I missing something here? This is a bit confusing.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -86,5 +134,5 @@ public PreparedQuery queryInDatabase(
     return datastore.prepare(new Query(entityName).setFilter(userFilter));
   }
 
-  // TODO: make error handling conform with Richie's error handling
+  // TODO: The same GitHub account can be used with multiple emails :(
```
Richie78321, line 137: That's a good point. In the future you could implement a check to determine if any user already exists with a given GitHub ID.

src/main/webapp/user.js
```diff
@@ -1,43 +1,52 @@
...
+    ' Please try again later.');
+
+  return errorFormattedFetchRequest.then((response) => response.json())
       .then((user) => {
         return user;
-      })
-      .catch((error) => {
-        console.error(error);
-        return null;
       });
```
Richie78321, line 25: What is the purpose of this then block? It seems like it doesn't do anything.

src/main/webapp/user.js
```diff
@@ -1,43 +1,52 @@
...
 }
 
 /**
  * Posts a user to the servlet
  * @param {URLSearchParams} userParams parameters with the users information
+ * @return {?Promise} Returns a prepared post user fetch request or null if the
+ * fetch request could not be prepared.
```
Richie78321, line 32: This doesn't reflect the behavior of this function:
```suggestion
 * @return {Promise} Returns a prepared post user fetch request.
```

src/main/webapp/user.js
Obinnabii: That's a good point!

src/main/java/com/google/opensesame/servlets/MentorObject.java
Obinnabii: Good to Know, I was under the impression that you were using Longs, I had it set to String earlier. I'll change it back

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: I'm getting rid of that functionality as a whole to keep the Servlet Restful. Good catch

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: sure thing!

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: I actually deprecated this functionality in order to keep the Servlet Restful. Good catch!
```suggestion
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: sure thing!

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: will do! I was trying to avoid having to make variables that will only be used once.

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: Yeah it does. 
```suggestion
    // Build and save the User's datastore entity
```

src/main/java/com/google/opensesame/servlets/MentorObject.java
Obinnabii: I will take this into account with the UserObject model.

Richie78321: Thanks. Sorry this wasn't more clear

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: Done! It made stuff a lot easier to read.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -110,21 +139,14 @@ public PreparedQuery queryInDatabase(
...
-    Key userKey = KeyFactory.createKey(PersonBuilder.ENTITY_NAME, userID);
-    try {
-      datastore.get(userKey);
-    } catch (EntityNotFoundException e) {
-      return false;
-    }
-    return true;
+  public PersonEntity hasProfile(String userID) {
+    return ofy().load().type(PersonEntity.class).id(userID).now();
   }
```
Richie78321, line 151: Thanks for leaving this in here for my sake, although I think now it's accessible enough to be done anywhere without this helper function. Also this function isn't static, so it means one would have to instantiate the `UserServlet` class to use this.

src/main/java/com/google/opensesame/servlets/UserObject.java
```diff
@@ -0,0 +1,123 @@
...
+  private final String name;
+  private final String userID;
+  private final Boolean isMentor;
+
+  public UserObject(
+      String userID,
+      String gitHubID,
+      ArrayList<String> interestTags,
+      ArrayList<String> menteeIDs,
+      ArrayList<Long> projectIDs,
```
josephdhurst, line 28: Would it be possible to replace `PersonObject` and `MentorObject` with this class? If it would push back the timeline too much don't worry about it.

Obinnabii:
I moved over to the One object model last night but I had conflicts getting changes from this branch. I'm going to close this PR and Open another with updated code.

PR #100, opened 2020-07-16
Fix signup.js

Richie78321:
It seems that in a recent merge some of the signup.js functionality stopped working. This is a quick PR to remedy that.


billyeh:
Aha. Note that the linter caught the mistake on master. +1 to the linter :D

PR #102, opened 2020-07-16
Build tests

Obinnabii:
Implement new build tests. We have been having issues with code that doesn't compile finding its way into master. This should help fix that. Once this is merged, I will make it a required check that code will build.

### DIFF
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -29,7 +29,7 @@ jobs:
         token: ${{secrets.PAT_TOKEN}}
     - name: Cache local npm dependencies
       uses: actions/cache@v2
-        with:
+      with:
         path: ~/.npm
         key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
         restore-keys: |
--- a/.github/workflows/formatting.yaml
+++ b/.github/workflows/formatting.yaml
@@ -13,10 +13,12 @@ jobs:
         with:
           token: ${{secrets.PAT_TOKEN}} # Re-running Linter on auto commit
       # Recommended: latest versions of Google Java Format require JDK 11+
-      - uses: actions/setup-java@v1
+      - name: Use Java 11
+        uses: actions/setup-java@v1
         with:
           java-version: "11"
-      - uses: axel-op/googlejavaformat-action@v3
+      - name: Format Java Code
+        uses: axel-op/googlejavaformat-action@v3
         with:
           args: "--replace"
 
@@ -27,6 +29,10 @@ jobs:
     - uses: actions/checkout@v2
       with:
         token: ${{secrets.PAT_TOKEN}}
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
     - name: Cache local npm dependencies
       uses: actions/cache@v2
       with:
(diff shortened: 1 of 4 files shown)


.github/workflows/formatting.yaml
```diff
@@ -41,3 +54,34 @@ jobs:
...
+    - uses: actions/checkout@v2 
+    - name: Use Java 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: "8"
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
+    - name: Cache local npm dependencies
```
billyeh, line 70: This is getting intense for me here, haha. So the idea is, we only have to install node dependencies once between Linting above and Building/Testing here?

.github/workflows/formatting.yaml
```diff
@@ -41,3 +54,34 @@ jobs:
       if: ${{ github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository }} 
       with:
         repo-token: ${{secrets.GITHUB_TOKEN}}
+
+  Build:
```
billyeh, line 58:
```suggestion
  BuildAndTest:
```
Since it runs the tests too?

Obinnabii: Good catch!

.github/workflows/formatting.yaml
```diff
@@ -41,3 +54,34 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v2 
+    - name: Use Java 8
+      uses: actions/setup-java@v1
+      with:
+        java-version: "8"
+    - name: Use Node.js
+      uses: actions/setup-node@v1
+      with:
+        node-version: '12.x'
```
Obinnabii, line 69:
```suggestion
        node-version: '10.x'
```

billyeh, line 69: interesting that here it's 12.x and above it's 10.x, not sure why? Wondering in general if we have to keep this updated or if there's some value we can pull it from.

billyeh: Not a big deal though I suppose.

.github/workflows/formatting.yaml
Obinnabii: Yes! and It makes it so that if any Action that sets up Maven or NodeJS has run before in any PR that this PR is built on and none of the dependencies have changed, GitHub won't re-build them. I read about it [here](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#about-caching-workflow-dependencies)
It cuts down IC check time drastically.

billyeh:
> I won't say that I'm the master of this type of thing, but it'll definitely help to have build tests. Thanks!

We're all learning :D

The fun part is when you can start adding badges to the README. And we could change the settings to force the linter/builder to pass before merging code haha which would help a lot with the issues with broken code at head we've been seeing.

.github/workflows/formatting.yaml
Obinnabii: That was a mistake haha

Richie78321:
> The fun part is when you can start adding badges to the README. And we could change the settings to force the linter/builder to pass before merging code haha which would help a lot with the issues with broken code at head we've been seeing.

Oh yeah that will be fun. Excited to learn!

PR #103, opened 2020-07-16
Project breakout page

Richie78321:
Create the React project breakout page that contains more detailed information about a project, most importantly information about the mentors associated with a project.

* Add to the `ExpandedProject` component to display more information about the project.
* Create the `MentorCard` component to display information about mentors associated with a project.
* Moves the column and padding div out of the `ProjectPreview` component to allow for it to be used in contexts other than a list.


src/main/webapp/js/projects/react/components/ExpandedProject.js
```diff
@@ -1,4 +1,6 @@
+import TagList from './TagList.js';
 import {expandedProjectType} from '../prop_types.js';
+import MentorCard from './MentorCard.js';
```
billyeh, line 3: nit: can we order our imports a little? Maybe put this import up with the TagList one.

src/main/webapp/js/projects/react/components/ExpandedProject.js
```diff
@@ -12,12 +14,50 @@ export default function ExpandedProject(props) {
     );
   }
 
-  // TODO : Add expanded project frontend view.
-  // For now this is just the name of the project to ensure that everything
-  // works.
+  const project = props.project;
+  const projectTags = project.previewData.topicTags;
+  // Adds the primary language to the list of project tags. Should add
```
billyeh, line 19:
```suggestion
  // Adds the primary language to the list of project tags. TODO add
```

src/main/webapp/js/projects/react/components/ExpandedProject.js
```diff
@@ -12,12 +14,50 @@ export default function ExpandedProject(props) {
...
+              <h1 className="card-title emphasis">
+                {project.previewData.name}
+              </h1>
+              <TagList tags={projectTags} />
+            </div>
+          </div>
+          <div className="ml-1 emphasis">Mentors:</div>
+          <div className="d-flex flex-wrap mb-2">
+            {/* TODO : Use the mentor User ID as the key. Once it is
+                available. */}
```
billyeh, line 40: Just wondering, what are the benefits of using the mentor user ID as key? IIUC key in react is just for rendering something correctly and an index should be just fine. I suppose we could remove 1 very tightly scoped variable?

src/main/webapp/js/projects/react/components/MentorCard.js
```diff
@@ -0,0 +1,38 @@
...
+ * @return {React.Component} Returns the react component.
+ */
+export default function MentorCard(props) {
+  const mentor = props.mentor;
+
+  return (
+    <div className="card project-card h-100">
+      <div className="d-flex flex-column card-body">
+        <div className="d-flex justify-content-center mb-2">
+          <img className="mentor-picture" src={mentor.image} alt="" />
```
billyeh, line 16: nit: provide an `alt`? Good for accessibility.

src/main/webapp/js/projects/react/components/ExpandedProject.js
Richie78321: Good question! Using the index in the list as the ID works, and it is what React defaults to if you don't manually set keys, but it actually is considered [bad practice](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318). Here is a snippet from [React's official documentation](https://reactjs.org/docs/reconciliation.html#recursing-on-children):

> When children have keys, React uses the key to match children in the original tree with children in the subsequent tree.

React uses keys to keep track of component children such that it only needs to re-render the items that have changed. You can imagine that if the ordering changes (like for example the order is reversed) React assumes that nothing has changed because the keys (if they are indices) have stayed the same, and therefore nothing is re-rendered and you have a problem.

> The best way to pick a key is to use a string that uniquely identifies a list item among its siblings. Most often you would use IDs from your data as keys.

React's recommended way to solve this problem is to use some sort of ID that can uniquely identify the data being represented by the component, which is in this case the User's ID.

src/main/webapp/js/projects/react/components/MentorCard.js
Richie78321: That's a good point. I initially kept it blank as is convention when an image is considered "decorative," but now that I think about it it seems like it might be important for context. Will add

src/main/webapp/js/projects/react/components/ExpandedProject.js
Richie78321: I agree. Thanks for the keen eye on keeping things neat.

src/main/webapp/js/projects/react/components/ExpandedProject.js
billyeh: Sounds good, thanks for the details and links!!

PR #105, opened 2020-07-16
Projects objectify

Richie78321:
Change project datastore code to use Objectify. This significantly reduces the amount of code necessary to save and load the `ProjectEntity` from datastore.

* Remove `ProjectDatastore.java`, which contained datastore saving and loading functionality that is now handled by Objectify.
* Change the queries used in both the `ProjectPreviewsServlet` and `ProjectServlet` to use Objectify.

A big TODO moving forward is consolidating the `ProjectPreviewData` and `ProjectData` classes, as there is a lot of overlap between them.


src/main/java/com/google/opensesame/servlets/ProjectPreviewsServlet.java
```diff
@@ -21,10 +20,9 @@
 
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    Query projectsQuery =
-        new Query("Project").addSort(ProjectEntity.NUM_MENTORS_KEY, SortDirection.DESCENDING);
+    List<ProjectEntity> projectEntities =
+        ofy().load().type(ProjectEntity.class).order("-numMentors").list();
```
Obinnabii, line 24: I think it would make sense to make strings into constants. So stuff like ```"numMentors"``` if it is used in multiple places. This makes changing the actual string easier in the event that we need to change it. I see that it isn't used anywhere else so this should be okay, just something to keep in mind.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -84,33 +65,28 @@ public ProjectEntity(
...
 
-  public int getNumInterestedUsers() {
-    return this.interestedUserIds.size();
+  @OnLoad
+  protected void checkNullLists() {
+    if (this.mentorIds == null) {
+      this.mentorIds = new ArrayList<String>();
+    }
+    if (this.interestedUserIds == null) {
+      this.interestedUserIds = new ArrayList<String>();
```
Obinnabii, line 81: I think Objectify throws an error if you save null. It might be worth it to use this check on save too.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
```diff
@@ -42,7 +43,7 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       return;
```
Obinnabii, line 43: I see that you used the ```getParameter``` function in your ```doGet```. does it do anything in this case?

src/main/java/com/google/opensesame/servlets/ProjectPreviewsServlet.java
Richie78321: I agree with you, although keep in mind that this string corresponds with the field name in the `ProjectEntity` class, so making a constant in this case might just be moving the problem somewhere else.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
Richie78321: Good point haha. It doesn't really in this case, does it? I'd like to keep it there for the time being in the case that I add more parameters down the line that might need different behaviors, and in that case I'd like things to stay consistent with this `getParameter` as well. Good catch

src/main/java/com/google/opensesame/projects/ProjectEntity.java
Richie78321: Oh it does? Let me verify that on my end, and in that case I'll add that to the `@OnSave`.

Richie78321: I just tested that and it seems like Objectify isn't having any issues "storing" null values. This corresponds with the idea that Objectify isn't actually "storing" these null and empty lists, but is instead ignoring them which causes them to be null on load.

src/main/java/com/google/opensesame/projects/ProjectEntity.java
```diff
@@ -84,33 +65,28 @@ public ProjectEntity(
...
-   * @param entity The Datastore Entity to add this ProjectEntity's properties to.
-   */
-  public void fillEntity(Entity entity) {
-    entity.setProperty(REPOSITORY_ID_KEY, repositoryId);
-    entity.setProperty(MENTOR_IDS_KEY, mentorIds);
-    entity.setProperty(INTERESTED_USER_IDS_KEY, interestedUserIds);
-    entity.setProperty(NUM_MENTORS_KEY, getNumMentors());
-    entity.setProperty(NUM_INTERESTED_USERS_KEY, getNumInterestedUsers());
+  public int getNumMentors() {
+    return this.mentorIds.size();
```
billyeh, line 86: Are this function and `getNumInterestedUsers` used anywhere? If not, maybe we can skip them for now.

I'm not 100% sure that they're better than just having the user call `ProjectEntity.mentordIds.size()` and there's no use-case currently.

Richie78321: That's a good point. `getNumMentors()` is currently being used in [`ProjectPreviewData.fromProjectEntity()`](https://github.com/googleinterns/step22-2020/blob/a912657e1f7f3b54f324e0b2810e16d544292e1f/src/main/java/com/google/opensesame/projects/ProjectPreviewData.java#L37) but I'm surprised I didn't notice how unnecessary these are because the lists themselves are public. I'll remove these.

PR #106, opened 2020-07-16
Finish has profile check in auth servlet

Richie78321:
Finish the `hasProfile` check to in the auth servlet using the `PersonEntity` objectify.


Obinnabii:
These changes were branched off a bad branch, I'll add them to the UserObject branch

PR #111, opened 2020-07-17
Add authenticated-only links to Navbar

Richie78321:
Add links that only display in the Navbar if a user is authenticated. **This is entirely for user experience and in no way literally prevents an unauthorized user from accessing a particular page**, so additional checks must be put in place for sensitive pages.

* Extract the fetch request for auth data from the `AuthButton` component so that it can use the `DataFetcher` component and so that the Navbar can use the authentication data as well without having to make a second fetch request.
* Add `requiresAuth` boolean to the Navbar URLs so they can be conditionally rendered.
* Revise the existing tests to support the new extracted auth data fetching and add new tests.


src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -2,59 +2,86 @@ import React from 'react';
...
   {
     href: '/projects.html',
     name: 'Projects',
+    requiresAuth: false,
+  },
+  {
+    href: '/dashboard.html',
+    name: 'Dashboard',
+    requiresAuth: true,
   },
```
Sami-2000, line 21: Is there a reason you don't include the mentor search page in this test?

src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -2,59 +2,86 @@ import React from 'react';
...
+    render(
+        <Navbar urls={navbarUrls} loading={false} authData={authData} />);
+
+    expect(screen.queryByText('Dashboard')).toBeNull();
+  });
+
+  it('does not display the restricted link while loading auth data', () => {
+    render(<Navbar urls={navbarUrls} loading={true} />);
+
+    expect(screen.queryByText('Home')).not.toBeNull();
```
Sami-2000, line 83: Here you explicitely check that 'Home' and 'Projects' are not null, but you do not check this for the unauthenticated or profile-less tests. I'm just wondering why.

src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -2,59 +2,86 @@ import React from 'react';
...
-    headers: {
-      'Content-Type': 'application/json',
-    },
-  });
-});
-
-afterAll(() => {
-  fetchMock.restore();
-});
+const mockAuthData = {
```
Sami-2000, line 24: Naming is slightly confusing here because you define other mock auth data later that you just call 'authData.'
Maybe if you can imply with the name that this is the authdata for an authorizd user with a profile, or even just 'defaultAuthData'?

src/main/webapp/js/common/react/tests/Navbar.test.js
Richie78321: Not really, no. Joe commented on this as well in the MVP branch, so I'll be releasing an additional PR that revises this to use the data directly from `display_navbar.js`

src/main/webapp/js/common/react/tests/Navbar.test.js
Richie78321: That's a good point, and seems to be a bit of an inconsistency. I'll update that now.

src/main/webapp/js/common/react/tests/Navbar.test.js
Richie78321: That's a good point! Will update, thanks!

PR #108, opened 2020-07-17
Prepare mentor search frontend for new user object.

Sami-2000:
Add profile pic, email, and location fields.
Allow for any aspect of a mentor to be null.
Use mailto to send emails to perspective mentors.

PR #121, opened 2020-07-20
Bulk Merge MVP Changes

Obinnabii:
### Obi's Changes
- Merge the Mentor and Person Objects into one ```UserData``` Class. 
- Merge the Mentor and Person Entities into one ```UserEntity```.
- Add an ```email``` field to the ```PersonObject``` and the dashboard.
- Deprecate the ```PersonBuilder```.
- Show projects that a user is involved with in the ```dashboard```.
- Remove unused code from previous iterations of the ```dashboard```. 
### Richie's Changes
* Implement the `hasProfile` check in `AuthServlet.java`
* Get the `UserData` of users associated with a project in `ProjectData.java`. This is replacing mock data.
* Move the `key` value to the parent divs of `ProjectPreview` components and `MentorCard` components.
Note: Changes to Navbar files `AuthButton.js`, `Navbar.js`, `NavbarLink.js`, `display_navbar.js`, `navbar_prop_types.js`, `Navbar.test.js`, and `NavbarLink.test.js` **are covered in #111 and don't need to be reviewed**
### Sami's Changes
- Use Objectify to overhaul the Mentors servlet
     - Post function to sign up to mentor can now create or modify project entities
     - New function to add a mock mentor to the page for testing and presentation
- Updated the Mentor search page frontend (some parts already approved in PR #108)
    - Check that each user field exists before adding it to the page
    - Link project tags to project breakout pages
    - Call the projects servlet to get the names of mentor-associated projects
    - Now we have 2 fully operation functions to populate the mentor search page, one with all real data, one with all mock


src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -2,59 +2,86 @@ import React from 'react';
...
     href: '/projects.html',
     name: 'Projects',
+    requiresAuth: false,
+  },
+  {
+    href: '/dashboard.html',
+    name: 'Dashboard',
+    requiresAuth: true,
   },
 ];
```
josephdhurst, line 22: is there a way we can reuse the urls variable from display_navbar.js?

josephdhurst: Also, why don't we call `display_navbar.js` `navbar.js`? It seems like display may be redundant (or I just don't get how it is different than all of the other javascript files in this codebase that ultimately display html).

src/main/webapp/js/common/react/tests/NavbarLink.test.js
```diff
@@ -3,19 +3,25 @@ import NavbarLink from '../components/NavbarLink.js';
...
+    render(<NavbarLink url={mockUrl} />);
     expect(screen.queryByText('Home')).not.toBeNull();
   });
 
   it('contains the intended href', () => {
-    render(<NavbarLink href="/home.html" name="Home" />);
+    render(<NavbarLink url={mockUrl} />);
     expect(screen.queryByText('Home')).toHaveAttribute('href', '/home.html');
   });
 });
```
josephdhurst, line 27: why isn't this a part of navbar.test.js?

src/main/webapp/mentors.js
```diff
@@ -12,7 +15,57 @@ function initForm() {
...
+      name: 'Obi Abii',
+      location: 'Idk maybe near Cornell',
+    },
+  ];
+  const mentorsContainer = document.getElementById('mentors-container');
+  if (!mentorsContainer) {
+    return;
+  }
+  mentorsContainer.innerHTML = '';
+  for (const mentor of mentors) {
```
josephdhurst, line 61: do we need `getMockMentors` anymore now that all of the data is being stored/retrieved from datastore?

src/main/webapp/mentors.js
```diff
@@ -79,7 +140,66 @@ function createMentorElement(mentor) {
...
+  mentorLocation.className = 'card-subtitle text-muted text-center mb-1';
+  mentorLocation.innerText = location;
+  return mentorLocation;
+}
+
+/**
+ * Creates an image for mentor card.
+ * @param {String} img
+ * @return {HTMLElement} mentorImg
+ */
```
josephdhurst, line 170: I think the function names here could be more specific, for example `createLocationHeader`.

src/main/webapp/mentors.js
```diff
@@ -79,7 +140,66 @@ function createMentorElement(mentor) {
 }
 
 /**
- * Creates a github profile button element from username.
+ * Creates an email for mentor card.
```
josephdhurst, line 143: Nit: fix this in all of the comments that mention mentor card.
```suggestion
 * Creates an email for a mentor card.
```

src/main/webapp/mentors.js
```diff
@@ -101,37 +221,62 @@ function createGitHubLink(gitHubID) {
...
+          const projectElement = document.createElement('a');
+          projectElement.href =
+              new URL('/projects.html#/' + projectID, window.location.origin);
+          projectElement.className = 'card-holder border border-muted' +
+              ' text-muted mr-1 mb-1 badge text-center';
+          const cardTitleElement = document.createElement('h4');
+          cardTitleElement.className = 'card-title dark-emph';
+          cardTitleElement.innerText = data.previewData.name;
+          projectElement.appendChild(cardTitleElement);
+          projectsDiv.append(projectElement);
```
josephdhurst, line 265: this is pretty long, maybe make this a separate function?

src/main/webapp/mentors.js
```diff
@@ -146,7 +291,8 @@ function submitForm(e) {
     body: encodedBody,
   });
 
-  standardizeFetchErrors(fetchRequest, 'Fetch failed message here', 'Server error message here').then((response) => {
+  standardizeFetchErrors(fetchRequest, 'Fetch failed message here',
+      'Server error message here').then((response) => {
     console.log('valid url');
     window.location.href = '/dashboard.html';
   }).catch((errorResponse) => {
```
josephdhurst, line 298: Fix ESLint errors

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -20,56 +24,49 @@
...
+    ArrayList<String> interests = new ArrayList<String>();
+    interests.add("skateboarding");
+    ArrayList<String> mentees = new ArrayList<String>();
+    ArrayList<String> projects = new ArrayList<String>();
+    projects.add(testRepoID.toString());
+    UserEntity mockMentor =
+        new UserEntity(id, "Sami-2000", interests, "samialves@google.com", projects, mentees);
+    ofy().save().entity(mockMentor);
+  }
+
```
josephdhurst, line 68: If I remember correctly, in the demo someone said that we can now add people to the datastore just by them signing up. If so can we delete this function?

src/main/webapp/user.js
```diff
@@ -4,25 +4,22 @@ import {
...
       fetchRequest,
       'Failed to communicate with the server. Please try again later.',
       'An error occcured while retrieving this account.' +
     ' Please try again later.');
 
-  return errorFormattedFetchRequest.then((response) => response.json())
-      .then((user) => {
-        return user;
-      });
+  return errorFormattedFetchRequest.then((response) => response.json());
```
josephdhurst, line 22: combine this into one line so that it matches the style used elsewhere.

src/main/webapp/signup.js
```diff
@@ -161,8 +143,30 @@ function createSignupBody() {
...
+    body: userParams,
+  });
+
+  const errorFormattedFetchRequest = standardizeFetchErrors(
+      fetchRequest,
+      'Failed to communicate with the server. Please try again later.',
+      'An error occcured while creating your account.' +
+    ' Please try again later.');
+
+  return errorFormattedFetchRequest;
```
josephdhurst, line 169: Don't store the result of a function call in a variable if it will just be returned in the next line. Just return the result of the function without storing it in a variable first. This is fewer lines of code.
```suggestion
   return  errorFormattedFetchRequest = standardizeFetchErrors(
      fetchRequest,
      'Failed to communicate with the server. Please try again later.',
      'An error occcured while creating your account.' +
    ' Please try again later.');
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -20,51 +15,41 @@
...
-
   @Override
-  // Instantiate datastore
-  public void init() {
-    datastore = DatastoreServiceFactory.getDatastoreService();
-  }
-
-  @Override
-  // Get a specific user. Return null if not found.
+  // Get a specific user. Return the currently signed-in user if no userId is
```
josephdhurst, line 25:
```suggestion
  // Get a user by their userId. Return the currently signed-in user if no userId is
```

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
   @Override
   // Send a user to datastore. Update the current information about the user if one exists.
   public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
-    String userGitHubID = request.getParameter("githubID");
-    if (userGitHubID == null || userGitHubID.trim().length() == 0) {
+    String userGitHubAuthToken = request.getParameter("gitHubAuthToken");
```
josephdhurst, line 61: nit: this can just be called `gitHubAuthToken`. the `user` part of the name is implied, because there is almost no scenario where a servlet should (or even could) have the auth token for a github account that doesn't belong to the current user.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-          "There is no GitHub ID affiliated with this request");
+          "GitHub OAuth token was never supplied",
+          HttpServletResponse.SC_UNAUTHORIZED,
+          "User could not be authenticated by GitHub, please try again");
       return;
     }
-    String[] userTags = request.getParameterValues("tags");
-    ArrayList<String> userTagsList =
-        userTags == null ? new ArrayList<String>() : new ArrayList<>(Arrays.asList(userTags));
+    String[] interestTagsParam = request.getParameterValues("interestTags");
```
josephdhurst, line 70: nit: remove `param` because:
 1. it doesn't give any extra information about what this variable does
 2. IIUC the data stored in `interestTagsParam` is the value for the "interestTags" parameter, so if we want to keep the suffix (which suggest against), we should change `Param` to `Value`

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+          HttpServletResponse.SC_UNAUTHORIZED,
+          "User could not be authenticated by GitHub, please try again");
+      return;
     }
-    return true;
+    // Build and send the User's datastore entity
+    ofy().save().entity(new UserEntity(userId, userGHMyself.getLogin(), userTags, user.getEmail()));
+    // TODO: add check for hasProfile.
   }
+  // TODO: make function to get users
```
josephdhurst, line 106: I think there is already a function to get a user based on ID. Either this TODO is already completed or it should be more specific.

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+              + e.getStackTrace()
+              + "Could not get authenticated User with the given token",
+          HttpServletResponse.SC_UNAUTHORIZED,
+          "User could not be authenticated by GitHub, please try again");
+      return;
     }
-    return true;
+    // Build and send the User's datastore entity
+    ofy().save().entity(new UserEntity(userId, userGHMyself.getLogin(), userTags, user.getEmail()));
+    // TODO: add check for hasProfile.
```
josephdhurst, line 104: Is this TODO necessary? Would it add significant latency? Would it just be sufficient to add .now() to the end of the `ofy().save` call?

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-   * @param value the object that will be compared to @param field
-   * @param operator the type of comparison to be made.
-   * @return {PreparedQuery} a Datastore prepared query.
-   */
-  public PreparedQuery queryInDatabase(
-      String entityName, String field, Object value, FilterOperator operator) {
-    Filter userFilter = new FilterPredicate(field, operator, value);
-    return datastore.prepare(new Query(entityName).setFilter(userFilter));
-  }
+    String userId = user.getUserId();
```
josephdhurst, line 86: create this variable closer to where it is used

src/main/java/com/google/opensesame/servlets/UserServlet.java
```diff
@@ -73,58 +58,50 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       return;
     }
-    String[] userTags = request.getParameterValues("tags");
-    ArrayList<String> userTagsList =
-        userTags == null ? new ArrayList<String>() : new ArrayList<>(Arrays.asList(userTags));
+    String[] interestTagsParam = request.getParameterValues("interestTags");
+    if (interestTagsParam == null) {
+      interestTagsParam = new String[] {};
+    }
+    ArrayList<String> userTags = new ArrayList<>(Arrays.asList(interestTagsParam));
```
josephdhurst, line 74: Create userTags close to where it is used.

src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -0,0 +1,137 @@
...
+  private final String location;
+  private final String name;
+  private final String userID;
+  private String email;
+
+  /**
+   * Create a UserData from a given UserEntity
+   *
+   * <p>Note: Emails will only ever be stored in a UserObject if the current user is authorized with
+   * the AuthServlet.
```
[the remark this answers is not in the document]
Obinnabii: We don't want to share user emails with unauthorized people to avoid spam and the misuse of our user's information by data scrapers. I'll update the Javadoc to reflect this reasoning.

src/main/java/com/google/opensesame/user/UserEntity.java
```diff
@@ -0,0 +1,123 @@
...
+
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Id;
+import com.googlecode.objectify.annotation.Index;
+import com.googlecode.objectify.annotation.OnSave;
+import java.util.ArrayList;
+
+/**
+ * The User Entity is designed to directly interface with datastore using Objectify. It holds the
+ * neccesary information to generate a UserObject.
```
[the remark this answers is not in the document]
Obinnabii: That makes a lot of sense! The XData classes are mainly used for serialization when sending JSON Objects. I also feel like the current names aren't as descriptive as I would like them to be. I'll work on changing these.

src/main/java/com/google/opensesame/user/UserEntity.java
```diff
@@ -0,0 +1,123 @@
...
+  public UserEntity(UserData userData) {
+    gitHubId = userData.getGitHubID();
+    interestTags = userData.getTags();
+    email = userData.getEmail();
+    userId = userData.getUserID();
+    projectIds = userData.getProjectIDs();
+    menteeIds = userData.getMenteeIDs();
+  }
+
+  // TODO: Cut down the number of constructors.
```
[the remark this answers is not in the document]
Richie78321: @Obinnabii Is there a reason why we couldn't just pass in null or an empty list for the `projectIds` and `menteeIds` to avoid having these extra constructors? We could handle null however we wanted (including creating empty lists when the inputted value is null).

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: It might not be necessary to have the TODO. There is already a check for whether a user has a profile in the frontend and it might add latency.

src/main/webapp/js/common/react/tests/Navbar.test.js
Richie78321: As for why is isn't called `navbar.js`, I wanted to avoid ambiguity between a `navbar.js` file and the `Navbar.js` file, where `Navbar.js` is the file containing the `Navbar` React component. `display_navbar.js` also makes it clear that it is not a React component and instead accomplishes the task of displaying the Navbar. I personally prefer `display_navbar.js`, but I'd be happy to hear more about why you might think otherwise!

Richie78321: And for reusing the URLs available in `display_navbar.js`, that's totally doable, and I see how that could reduce the number of experimental variables present if the tests ever fail. I'll update that.

src/main/webapp/js/common/react/tests/NavbarLink.test.js
Richie78321: This is testing the specific functionality of the `NavbarLink` component compared to the functionality of the entire `Navbar` component. I thought it would be best to put the tests relating to the `NavbarLink` in `NavbarLink.test.js` instead of in `Navbar.test.js` to keep things organized. I realize the confusion might be coming from the fact that these tests are somewhat redundant, as the `Navbar` tests indirectly test this functionality as well, and would fail if this functionality did not work. Is that a sign to remove these tests?

src/main/java/com/google/opensesame/servlets/UserServlet.java
Obinnabii: The main aim of this naming scheme was to differentiate between the ```String[]``` gotten from the request and the ```ArrayList``` used in building a ```UserEntity``` . but I see what you are saying. I'll avoid creating two variables entirely by casting to an ```ArrayList``` when I build the ```UserEntity```

src/main/java/com/google/opensesame/user/UserEntity.java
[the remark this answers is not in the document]
Obinnabii: my main issue with passing in null values is that it is not good practice. The standard with Java is to avoid using ```null``` as much as possible. It would also make the documentation a bit more confusing. This way, you always know what you have to put in.

__Possible Solutions__
1. Forgo a constructor entirely. The class can be instantiated without a constructor and all the fields can be set because they are public. 
_con_
- The main shortcoming of this method is that there are some fields we want to ensure always gets set. My current solution works so that you always instantiate a ```UserEntity``` object with the required fields and any optional fields are set as empty lists. 
2. A builder. This might work. 
_con_
- I am reluctant to add more classes to the 2 that we have now. and It would also make allow the compulsory fields to be left unset.

3. Leave the constructors as we have them. 
_con_
- we currently have 4 constructors.

4. Create the ```UserEntity``` Object without the optional fields and set those fields after construction.
_con_
- It seems hacky and construction of objects could take multiple lines depending on the situation.

PR #126, opened 2020-07-22
Consolidate to single project data class and servlet

Richie78321:
Consolidate `ProjectPreviewData` into `ProjectData` and `ProjectPreviewServlet` into `ProjectServlet`.

* Moves `ProjectData` creation from static factory functions to public methods in each `ProjectData` instance.
* The `ProjectServlet` can now respond to all queries for projects, and it now includes options to select the amount of data needed: the name, the preview data, or all of the data.
  * If a list of IDs is supplied, the servlet will respond with those projects. Otherwise, it will respond with all of the projects in a list ordered by the number of mentors descending.
  * As more search functionality is implemented like ordering, filtering, and pagination, more functionality will be added to the non-ID queries.
* Updated uses of the servlet in `dashboard.js` and `mentors.js`.
* Updated uses of the servlet in the project React components.

*Feel free to give me feedback on how I went about implementing the `ProjectData` class, and let me know if you think there is a better design pattern I could use!*


src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -10,57 +10,105 @@
...
-   * @return Returns the created ProjectData.
-   * @throws IOException Throws when there is an error communicating with the GitHub API.
+   * @return Returns the ProjectData instance with the data assigned.
+   * @throws IOException Throws if there is an error communicating with the GitHub API.
    */
-  public static ProjectData fromProjectEntity(ProjectEntity projectEntity, GHRepository repository)
-      throws IOException {
-    ProjectPreviewData previewData =
-        ProjectPreviewData.fromProjectEntity(projectEntity, repository);
+  public ProjectData withFullData() throws IOException {
```
Sami-2000, line 102: Q: Having all of these constructor-helpers (withName, withPreviewData,withFullData) is definitely an efficient way to send the minimum necessary data. However, considering how large this file has now gotten, idk if this may be considered premature optimization? (honestly don't know)

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
```diff
@@ -7,54 +7,117 @@
...
           HttpServletResponse.SC_BAD_REQUEST,
-          "Failed to retrieve project data.");
-      return;
+          "Failed to interpret the request. Please try again later.");
+    }
+
+    String[] projectIds = request.getParameterValues(PROJECT_ID_PARAM);
+    Collection<ProjectEntity> projectEntities;
+    if (projectIds != null) {
+      // Get projects by ID if IDs were specified.
```
Sami-2000, line 102: Not sure this comment is necessary. Your variable names and logic here are already telling me what you are doing.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
```diff
@@ -7,54 +7,117 @@
...
+            response,
+            "One or more of the supplied IDs is invalid and those projects could not be found.",
+            HttpServletResponse.SC_NOT_FOUND,
+            "One or more of the requested projects does not exist. Please try again later.");
+      }
+    } else {
+      // TODO(Richie): Add parameters for ordering and filtering the query.
+      // TODO(Richie): Add pagination support.
+      // Get all projects in default ordering if no IDs are specified.
+      projectEntities = ofy().load().type(ProjectEntity.class).order("-numMentors").list();
```
Sami-2000, line 116: Love that u at least order by numMentors for now!

src/main/webapp/mentors.js
```diff
@@ -237,8 +237,8 @@ function getProjectName(projectID) {
       'Encountered a server error, please try again later.')
       .then((response) => response.json()).then((data) => {
         console.log('got project data');
-        console.log(data.previewData.name);
-        return data.previewData.name;
+        console.log(data[0].name);
```
Sami-2000, line 240: Thanks for going ahead and changing this so it doesn't break!

Obinnabii: Oh cool, dashboard.js also uses previewData, could you maybe make the change there too?

src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -10,57 +10,105 @@
...
+  public List<String> topicTags;
+  public String primaryLanguage;
+  public Integer numMentors;
+  public String repositoryId;
+  public List<UserData> mentors;
+
+  /**
+   * Create a ProjectData object from a ProjectEntity and its associated GitHub repository.
+   *
+   * <p>The main use of this constructor is for testing, where the GitHub repository must be mocked.
```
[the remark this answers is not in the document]
Richie78321: It seems to be common practice in Javadoc. Take a look at this [random file from Objectify](https://github.com/objectify/objectify/blob/master/src/main/java/com/googlecode/objectify/LoadResult.java) I opened. The tags were added by the linter, so I'm not exactly sure when they do and do not apply.

src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -10,57 +10,105 @@
...
 public class ProjectData {
+  private transient GHRepository repository;
+  private transient ProjectEntity projectEntity;
+  public String name;
+  public String description;
+  public List<String> topicTags;
+  public String primaryLanguage;
+  public Integer numMentors;
+  public String repositoryId;
+  public List<UserData> mentors;
```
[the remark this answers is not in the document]
Richie78321: Yes, I'd be happy to. I have some solid opinions on why this should be `UserData` and why in the case of the `UserData` class it should also be `ProjectData`.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
```diff
@@ -7,54 +7,117 @@
...
+   * @throws IOException Throws if there was an error retrieving the GitHub repository associated
+   *     with a ProjectEntity.
+   */
+  private static Collection<ProjectData> createProjectData(
+      Collection<ProjectEntity> projectEntities, boolean getFullData, boolean getPreviewData)
+      throws IOException {
+    List<ProjectData> projectData = new ArrayList<ProjectData>();
+    for (ProjectEntity projectEntity : projectEntities) {
+      ProjectData newProjectData = new ProjectData(projectEntity);
+      // TODO(Richie): Is there a better way to do this than an if-else block?
```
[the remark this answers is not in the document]
Richie78321: That's a good point; I was more thinking about how I could design it such that the if-else block could be avoided altogether, but I'll definitely take another look at this code and see if the constructor idea fits better.

src/main/java/com/google/opensesame/projects/ProjectData.java
Richie78321: I personally don't think so. I'd be happy to hear from either Bill or Joe for their perspectives, but knowing that we are using each one of these types of data (`withName` on the mentor and dashboard pages, `withPreviewData` on the search page, and `withFullData` on the project breakout page), it doesn't seem premature.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
Richie78321: Good point, will remove. Thanks!

src/main/webapp/mentors.js
Richie78321: @Obinnabii Do [these changes](https://github.com/googleinterns/step22-2020/pull/126/files#diff-d5bd38389c6679a39a26096d66d07de0) cover it?

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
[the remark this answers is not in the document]
billyeh: My 2¢: the bools are not good, I agree.

How about this: ProjectData just has separate getters for description, topicTags, etc. If we want to do caching there, we can have private class variables initialized to null and only get the data from Github when they're null.

Instead of passing booleans to one function, we can have `createProjectPreviewData`, `createProjectFullData`, and `createProjectData` functions that just get the relevant fields.

And actually looking at the boolean logic below I feel it'd be simpler to just have different servlets/endpoints for each of the above methods.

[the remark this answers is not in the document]
Richie78321: > How about this: ProjectData just has separate getters for description, topicTags, etc. If we want to do caching there, we can have private class variables initialized to null and only get the data from Github when they're null.

Keep in mind this object is being prepared for serialization and not much else, so the main intention is to assign the values to the class variables before the object is serialized to JSON.

> Instead of passing booleans to one function, we can have createProjectPreviewData, createProjectFullData, and createProjectData functions that just get the relevant fields.

This sounds very similar to what has already been implemented, so please let me know if I'm not understanding it completely.

> And actually looking at the boolean logic below I feel it'd be simpler to just have different servlets/endpoints for each of the above methods.

This would definitely remove the need for an if-else block anywhere in the code. I'm thinking abstracting the `ProjectEntity` query system to avoid repeat and using the different endpoints to implicitly determine what data is needed could work!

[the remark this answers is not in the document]
Richie78321: Unless there are any objections I'm definitely going to move forward with the different endpoints idea, thanks Bill and Sami!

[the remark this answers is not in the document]
billyeh: Ah yeah actually I wasn't sure about the purpose of this class, even though I read the comment in ProjectData, line 15. Maybe just say "This is a read-only class is for serializing data from Datastore to JSON" - certain buzzwords like "read-only" convey a lot. 

So yeah the logic can be even simpler then, no null-checking, but I still think there should be getters.

IMO it's not a good assumption to make that e.g. `withFullData` will always call `withPreviewData` which will always call `withName`, I feel that could easily change in the future on a whim depending on what we want to throw on the pages. 

I think it's best to be able to just specify what you want. Currently what we have is similar to an inheritance-based approach which requires a lot of layers of looking at what data this function gets, which calls this function which calls this function, etc. and isn't very flexible. Maybe that's the premature optimization that was referred to?

src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -10,57 +10,105 @@
...
 import org.kohsuke.github.GHRepository;
-import org.kohsuke.github.GitHub;
 
 /**
- * A class containing all relevant data for a project. This is intended to be serialized and sent to
- * the client.
+ * A class containing the data for a project. This is intended to be serialized and sent to the
+ * client. A ProjectData object can define as many or as few of the fields as necessary depending on
+ * the use case, as undefined values will not be serialized to JSON and will therefore not be sent
+ * to the client.
```
billyeh, line 18: Added a comment below.

This information is useful, but not sure if it's the first thing to write in class documentation. It's kind of an implementation detail. As we've discussed, it's important to clearly define the use-cases i.e. "why" would I use this class/what would I use it for?

billyeh: A stab at an example docstring: "Read-only utility class for converting a ProjectEntity datastore object into a POJO for JSON serialization" 

:) something like that haha sorry for throwing in the POJO jargon.

src/main/java/com/google/opensesame/servlets/ProjectServlet.java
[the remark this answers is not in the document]
Richie78321: Yeah that makes a lot of sense when you say it that way. I'll revise based on those recommendations, thanks so much!

src/main/java/com/google/opensesame/projects/ProjectData.java
Richie78321: I've seen POJO being thrown around before in Objectify, so no worries there! I'll update this to reflect more of the "why" when I push my other changes.

PR #128, opened 2020-07-22
Homepage

Sami-2000:
Add a sign-up button,  'about us' section, and 3 basic user flows to the homepage.


Richie78321:
Hey Sami, it seems like the commit history is a bit messed up on this branch. Consider un-committing your changes with
```
git reset 8d212c0
```
and then stashing the changes, moving to a branch based off of master, and then re-committing your changes from the stash. You can then force-push to this remote branch to update the commit history. (I know force-pushing isn't ideal so if someone knows of a better solution go with that)

PR #129, opened 2020-07-22
Add authentication redirect and add to README

### LINKED ISSUE #101: Authentication redirect for signup form
It would be more useful for users to know that they need to authenticate before starting the signup form. Add an authentication check and sign-in redirect. Consider allowing the request to specific the redirect URL.

Richie78321:
Add authentication redirect and authentication data fetching to a module for easy access and to avoid repeated calls to the authentication servlet on the same page.

* Add information about the user signup flow to the README
* Create a `sign_in_redirect.html` page that redirects users after logging in with the Users API
* Add `auth_check.js` module that gets authentication data from the servlet and provides a helper function for redirecting based on that data.
* Change the Navbar to reuse the authentication data present in `auth_check.js`.
* Create `default_auth_redirect.js`, a convenience module that implements the default redirect behavior for an authenticated-only page.
* Add checks for no JavaScript to the redirect page and the projects page.

closes #101

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,5 +65,20 @@ Java unit testing is done with [JUnit 4](https://junit.org/junit4/).
 **To add another JS test**, simply add a JS file that ends with `.test.js` (ex: `script.test.js`) and it will automatically be run by Jest.
 **To add another Java test**, add a Java file to the testing directory `src/test/java/`. Be sure to follow the Java conventions for package directory structure, for example if you're testing a package in `com.google.opensesame.github`, the test should be placed in the directory `src/test/java/com/google/opensesame/github`.
 
+## User Authentication and Sign Up
+User signup consists of three parts:
+1. Website authentication (a user creates an account on our website)
+  * This is done using the [Users API](https://cloud.google.com/appengine/docs/standard/java/users)
+  * In the future this could be migrated to the [Firebase Admin SDK](https://firebase.google.com/docs/reference/admin) or similar OAuth-based authentication API such that GitHub authentication and website authentication could become a single step.
+2. GitHub authentication (a user verifies ownership of a GitHub account, which is then linked with the account on our website)
+  * This is done using [Firebase Authentication](https://firebase.google.com/docs/auth) with [GitHub OAuth](https://docs.github.com/en/developers/apps/about-apps#about-oauth-apps)
+  * An access token is granted to the user, which is then sent to the server during the profile creation step and verified on the backend. 
+  * Once the access token is verified, the user has proven ownership of that GitHub account, and it is then stored by ID in the website profile.
+3. Profile creation (a user with an authorized account can add more information about themselves)
+  * After a user completes website authentication, they are redirected to a form for profile creation.
+  * GitHub authentication is completed during profile creation.
+
+A user is only considered to have an account on Open Sesame if they have completed all three steps outlined above.
+
 ## File Structure
 Please refer to [this](https://maven.apache.org/guides/introduction/introduction-to-the-standard-directory-layout.html) resource to learn more about what each project directory should be used for, and take a look at [this section](https://maven.apache.org/guides/getting-started/#how-do-i-make-my-first-maven-project) of the Maven getting started guide to see how the project naming scheme affects directory structure.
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -65,5 +65,20 @@ Java unit testing is done with [JUnit 4](https://junit.org/junit4/).
...
+## User Authentication and Sign Up
+User signup consists of three parts:
+1. Website authentication (a user creates an account on our website)
+  * This is done using the [Users API](https://cloud.google.com/appengine/docs/standard/java/users)
+  * In the future this could be migrated to the [Firebase Admin SDK](https://firebase.google.com/docs/reference/admin) or similar OAuth-based authentication API such that GitHub authentication and website authentication could become a single step.
+2. GitHub authentication (a user verifies ownership of a GitHub account, which is then linked with the account on our website)
+  * This is done using [Firebase Authentication](https://firebase.google.com/docs/auth) with [GitHub OAuth](https://docs.github.com/en/developers/apps/about-apps#about-oauth-apps)
+  * An access token is granted to the user, which is then sent to the server during the profile creation step and verified on the backend. 
+  * Once the access token is verified, the user has proven ownership of that GitHub account, and it is then stored by ID in the website profile.
+3. Profile creation (a user with an authorized account can add more information about themselves)
```
Sami-2000, line 77: If, for now, "add more info about themselves" just means add interest tags, then I would probably rephrase this description.

src/main/webapp/js/auth_check.js
```diff
@@ -0,0 +1,68 @@
...
+}
+
+/**
+ * A promise that returns the authentication data once it has been fetched.
+ * This is useful if you need to define more advanced behavior with the
+ * authentication data.
+ * @type {Promise}
+ */
+export let authDataFetch = createAuthFetch();
+/**
```
Sami-2000, line 36: Nit: Add a blank line between these?

src/main/webapp/projects.html
```diff
@@ -14,6 +14,9 @@
   <body>
     <div id="navbar-container"></div>
     <div id="projects-container"></div>
+    <noscript>
+      Please enable JavaScript to use this website.
```
Sami-2000, line 18: Q: Should we be adding this check to every page in the site?

Sami-2000: (I'm pretty sure almost nothing we have would work without js)

src/main/webapp/js/auth_check.js
```diff
@@ -0,0 +1,68 @@
...
+ * Creates the authentication fetch request.
+ * @return {Promise} Returns the authentication fetch request.
+ */
+function createAuthFetch() {
+  const errorFormattedRequest = standardizeFetchErrors(
+      fetch(makeRelativeUrlAbsolute('/auth')),
+      'Failed to communicate with authentication server.',
+      'Failed to verify authentication.');
+  return errorFormattedRequest
```
Obinnabii, line 15: is there any reason for separating these?

README.md
Richie78321: I definitely see what you're saying, but I wonder if that just presents an opportunity for the documentation to become out of date? As it stands now it still conveys enough information but is general enough to still apply with future changes.

src/main/webapp/js/auth_check.js
Richie78321: For readability. Otherwise, the line wrapping gets pretty crazy, although if you think this is worth changing lmk.

src/main/webapp/projects.html
Richie78321: Yeah most likely. I wanted the choice to ultimately reside in the "owner" of the page, so feel free to add this to your HTML pages as well.

src/main/webapp/js/auth_check.js
```diff
@@ -0,0 +1,68 @@
...
+ * @param {string} [noAuthRedirect=null] Where to redirect if the user is not
+ *    authenticated. By default the user will not be redirected.
+ * @param {string} [authRedirect=null] Where to redirect if the user is
+ *    authenticated. By default the user will not be redirected.
+ * @param {string} [noProfileRedirect=null] Where to redirect if the user does
+ *    not have a profile. By default the user wil not be redirected.
+ * @return {Promise} Returns a promise for the authentication check.
+ */
+export function authRedirect(
+    noAuthRedirect = null, authRedirect = null, noProfileRedirect = null) {
```
Obinnabii, line 50: I think it would make sense to have all the functions that handle redirection in this file/ in the same file.

Richie78321: Are there any other redirection functions that are missing from this file? If you're referring to things like `sign_in_redirect.js` and `default_auth_redirect.js`, inevitably some file must call the function implemented in the module. If the module itself calls the function, then there would be no room for customization of the redirection function.

README.md
Richie78321: That's a good middleground, thanks! Will update.

PR #134, opened 2020-07-23
Add MailtoUI to Mentor Page Mentor card.

Sami-2000:
Add "mailtoui" class to the mentor card email links. Run mailtoUI once everything has been loaded onto the page.

### DIFF
--- a/src/main/webapp/mentors.js
+++ b/src/main/webapp/mentors.js
@@ -50,6 +50,7 @@ function getMentors() {
       console.log(mentor);
       mentorsContainer.appendChild(createMentorElement(mentor));
     }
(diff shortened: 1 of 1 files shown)


src/main/webapp/mentors.js
```diff
@@ -50,6 +50,7 @@ function getMentors() {
       console.log(mentor);
       mentorsContainer.appendChild(createMentorElement(mentor));
     }
+    mailtouiApp.run();
```
Richie78321, line 53: The linter doesn't seem to like this -- you might need to add it to the list of globals in the `.eslintrc` file (I'll probably have to do the same on my PR)

PR #136, opened 2020-07-24
Button

Sami-2000:
Make the homepage button dependent on user auth status. It will now display the same message and links as the button in the navbar.


src/main/webapp/index.html
```diff
@@ -10,7 +10,8 @@
 <div class="container">
   <h1 class="display-2 mb-5 mt-5 text-center">The Open Source Community is for <span class="emphasis">Everyone</span></h1>
   <!-- TODO: Change button text/link depending on user authorization. -->
```
Richie78321, line 12: Can we remove this now?
```suggestion
```

src/main/webapp/index.html
```diff
@@ -10,7 +10,8 @@
...
-  <div class="d-flex justify-content-center"><div class="btn btn-emphasis mt-5 mb-5 text-center btn-lg btn-block btn-main">Create a Profile!</div></div>
+  <!-- <div class="btn btn-emphasis mt-5 mb-5 text-center btn-lg btn-block btn-main">Create a Profile!</div> -->
```
Richie78321, line 13: Is there a reason this is still here? If not, can we remove it?
```suggestion
```

src/main/webapp/js/common/react/index.js
```diff
@@ -0,0 +1,21 @@
+import {DataFetcher} from './components/DataFetcher.js';
+import {AuthButton} from './components/AuthButton.js';
+import {authDataFetch} from '../../auth_check.js';
+
+
+function renderButton(dataFetcher) {
```
Richie78321, line 6: Can we fix this and below linter errors?

src/main/webapp/index.html
Obinnabii, line 13: could we remove this commented code?

src/main/webapp/index.html
```diff
@@ -10,7 +10,8 @@
...
+  <div id="button-container" class="d-flex justify-content-center"></div>
```
Richie78321, line 14: nit: It might be worth making this id a bit more specific (keep in mind you would need to change it in the JS below too)
```suggestion
  <div id="auth-button-container" class="d-flex justify-content-center"></div>
```

src/main/webapp/index.html
Sami-2000: Oh yes oc!

PR #138, opened 2020-07-24
Fix Duplicated ESLint Annotation

Obinnabii:
- Remove the [ESLint Action](https://github.com/marketplace/actions/eslint-action#eslint-action) GitHub Action that annotates ESLint errors and warnings in PRs. It turns out that the local linter does a similar thing and this has been causing duplicated annotation in the ```files changed``` tab
- Move from a ```continue-on-error: true```  GitHub Actions model to an ```if: always()``` model in the ESLint Action. Using ```continue-on-error: true``` makes GitHub believe that the ```ESLint``` Check passed when it failed. The ```if: always()``` model allows the ```ESLint``` GitHub Action to continue running even if it fails. The key difference is that the ```ESLint``` check is registered as a failure with ```if: always()``` but not with ```continue-on-error: true```.
- add ```firebase``` to the list of ```ESLint``` global variables.
- change ```ESLint``` run directory.


.eslintrc
```diff
@@ -19,5 +19,6 @@
     "PropTypes": "readonly",
     "ReactRouterDOM": "readonly",
     "mailtouiApp": "readonly",
+    "firebase": "readonly"
```
billyeh, line 22:
```suggestion
    "firebase": "readonly",
```
nit :)

PR #139, opened 2020-07-24
Small navbar fixes

Richie78321:
* Move the `AuthButton` into the collapsible div so it scales properly on mobile.
* Make the Navbar sticky.
* Removed the tests for the `NavbarLink` because they are redundant with the tests for the `Navbar`.
* Updated the urls in the `Navbar` tests to reflect that actual urls.


src/main/webapp/js/common/react/tests/Navbar.test.js
```diff
@@ -14,14 +14,19 @@ const navbarUrls = [
     name: 'Projects',
     requiresAuth: false,
   },
+  {
```
Sami-2000, line 17: Thanks for adding this!

Richie78321: Np! It's definitely better to have the test URLs match up with the production URLs anyways.

src/main/webapp/js/common/react/display_navbar.js
```diff
@@ -37,6 +37,7 @@ const urls = [
  */
 function initNavbar() {
   const navbarContainer = document.getElementById('navbar-container');
+  navbarContainer.classList.add('sticky-top');
```
billyeh, line 40: nice to learn about this class

Richie78321: It's super convenient, although it isn't yet [supported in all browsers](https://caniuse.com/#feat=css-sticky). The good news is that there's no real downside if the browser doesn't support it beyond the navbar not sticking to the top of the page.

PR #140, opened 2020-07-24
Add testing to projects frontend

Richie78321:
Add testing to the project preview list and project breakout page.

* Add basic mock for `mailtouiApp` for testing.
* Add `checkTesting()` to all of the new React components being tested.
* Add `inRouter` to ProjectPreview component to allow for it to be tested outside of a Router.
* Update the `mentorType` PropType to include all of the values used in the MentorCard.
* Move all new and existing JS tests to a single directory `src/main/webapp/js/tests`.
* Create `ExpandedProject` tests.
* Create `ProjectList` tests.


src/main/webapp/js/projects/react/prop_types.js
```diff
@@ -1,3 +1,6 @@
+import {checkTesting} from '../../checkTesting.js';
+checkTesting();
```
josephdhurst, line 2: nit: consider changing `checkTesting` to something that is more descriptive of what this function is doing. IMO something like `setupIfTesting`, `configureGlobalsForEnvironment` or `setGlobalsIfTesting` would be better.

josephdhurst: I'd expect that a function called `checkTesting` would return a boolean telling whether this was a test environment or not. I wouldn't expect it to modify the global state.

Richie78321: I totally agree. I'll change it to `setGlobalsIfTesting`.

PR #141, opened 2020-07-27
Get ProjectData from the UserServlet

Sami-2000:
Allow UserData objects to contain a list of ProjectData objects (following form of ProjectData).


src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,11 +26,14 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
+    addMockMentor(response);
```
Richie78321, line 29: Can we add a TODO here to remove this for production?

src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -15,8 +20,7 @@
 public class UserData {
   private Boolean isMentor;
   private final ArrayList<String> interestTags;
-  private final ArrayList<String> menteeIds;
-  private final ArrayList<String> projectIds;
+  private final ArrayList<ProjectData> projects = new ArrayList<ProjectData>();
```
Richie78321, line 23: Nice! Thanks for making this change.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,11 +26,14 @@
...
     List<UserEntity> mentorEntities = ofy().load().type(UserEntity.class).list();
     ArrayList<UserData> mentors = new ArrayList<UserData>();
     for (UserEntity entity : mentorEntities) {
       if (entity.isMentor()) {
-        mentors.add(new UserData(entity));
+        UserData mentorData = new UserData(entity);
+        mentorData.getProjects();
```
Richie78321, line 35: It seems like this is just a getter function. Is there a reason that this is being called here?

src/main/java/com/google/opensesame/user/UserEntity.java
```diff
@@ -35,8 +34,11 @@ public UserEntity(UserData userData) {
     interestTags = userData.getTags();
     email = userData.getEmail();
     userId = userData.getUserID();
-    projectIds = userData.getProjectIDs();
-    menteeIds = userData.getMenteeIDs();
+    projectIds =
+        new ArrayList<String>(
+            userData.getProjects().stream()
+                .map(x -> x.getRepositoryId())
```
Richie78321, line 40: nit: Could we use a more descriptive name here?
```suggestion
                .map(projectData -> projectData.getRepositoryId())
```

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
Obinnabii: good catch, that was a call to deprecated functionality

src/main/webapp/mentors.js
```diff
@@ -18,7 +18,8 @@ function initForm() {
  * User Object
  * @typedef {Object} User
  * @property {string[]} interestTags - The user's tags
- * @property {string[]} projectIDs - The user's projects
+ * @property {string[]} projectIDs - The user's project IDs
```
Obinnabii, line 21:
```suggestion
```

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,11 +26,13 @@
   @Override
   public void doGet(HttpServletRequest request, HttpServletResponse response)
       throws ServletException, IOException {
+    addMockMentor(response); // TODO: remove by production
     List<UserEntity> mentorEntities = ofy().load().type(UserEntity.class).list();
```
josephdhurst, line 30: IIUC this list doesn't only contain mentors (otherwise the `if (entity.isMentor())` statement is unnecessary. Rename `mentorEntities` to something else that conveys that this list also contains people who aren't mentors. A suggestion I have is to call it `userEntities`.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,11 +26,13 @@
...
     ArrayList<UserData> mentors = new ArrayList<UserData>();
     for (UserEntity entity : mentorEntities) {
       if (entity.isMentor()) {
-        mentors.add(new UserData(entity));
+        UserData mentorData = new UserData(entity);
+        mentors.add(mentorData);
```
josephdhurst, line 35: Rather than storing `UserData(entity)` in a variable then appending it to `mentors`, skip the step where you store it in a variable. This reduces 2 lines of code into 1.
```suggestion
        mentors.add(new UserData(entity));
```

src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -110,30 +113,24 @@ public String getUserID() {
...
-   * Add a project to the list of projects associated with this instance of UserData. Remember to
-   * store this change in datastore with a UserEntity.
+   * Add the {@code ProjectData} associated with the projects in {@code projectIds} to {@code
+   * this.projects}.
+   *
+   * @param projectIds
    */
-  public void addMentee(String menteeId) {
-    projectIds.add(menteeId);
+  private void generateProjects(ArrayList<String> projectIds) throws IOException {
```
josephdhurst, line 127: How about naming this `loadProjects` instead? The name `generateProjects` makes me thing that somehow there is logic in this function that creates the projects without using pre-stored data.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,13 +27,19 @@
...
+        ofy().load().type(UserEntity.class).filter("isMentor", true).list().stream()
+            .map(
+                userEntity -> {
+                  try {
+                    return new UserData(userEntity);
+                  } catch (IOException e) {
+                    return null;
+                  }
+                })
+            .collect(Collectors.toList());
```
Obinnabii, line 41: @josephdhurst does this fix your previous issue?

josephdhurst:
```suggestion
    List<UserEntity> mentorEntities = ofy().load().type(UserEntity.class).filter("isMentor", true).list();
    ArrayList<UserData> mentors = new ArrayList<UserData>();
    for (UserEntity entity : mentorEntities) {
      mentors.add(new UserData(entity));
    }
```

PR #142, opened 2020-07-27
Projects backend testing

Richie78321:
Add testing for querying the projects Datastore.

* Add the necessary dependencies for creating in-memory test instances of Datastore and Objectify.
* Add tests for `ProjectEntity.queryFromRequest` which is used by all of the project data servlets.
* Add tests for the `ProjectEntity.fromRepositoryIdOrNew` which is used by the Mentor's `doPost` servlet.

### DIFF
--- a/src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
+++ b/src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
@@ -2,6 +2,7 @@ package com.google.opensesame.projects;
 
 import static com.googlecode.objectify.ObjectifyService.ofy;
 import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
@@ -134,5 +135,27 @@ public class ProjectEntityTest {
         new ProjectEntity[] { firstMockProject, secondMockProject, thirdMockProject };
     assertNotNull(queryResult);
     assertArrayEquals(expectedQueryResult, queryResult.toArray());
+  }
+
+  @Test
+  public void fromRepositoryId() {
+    // Expect the function to return an existing ProjectEntity from the Datastore because an entity
+    // with the supplied ID already exists.
+
+    ProjectEntity projectEntity = 
+        ProjectEntity.fromRepositoryIdOrNew(firstMockProject.repositoryId);
+
+    assertEquals(firstMockProject, projectEntity);
+  }
+
+  @Test
+  public void fromNewRepositoryId() {
+    // Expect the function to return a new ProjectEntity because no entity with that supplied ID
+    // exists in the Datastore.
+
+    String newId = "newId";
+    ProjectEntity projectEntity = ProjectEntity.fromRepositoryIdOrNew(newId);
+
+    assertEquals(newId, projectEntity.repositoryId);
   }
 }
--- a/src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
+++ b/src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
@@ -1,0 +1,138 @@
+package com.google.opensesame.projects;
+
+import static com.googlecode.objectify.ObjectifyService.ofy;
+import static org.junit.Assert.assertArrayEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.anyInt;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.never;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
+import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
+import com.googlecode.objectify.ObjectifyFactory;
+import com.googlecode.objectify.ObjectifyService;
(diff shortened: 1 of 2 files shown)


src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
```diff
@@ -0,0 +1,161 @@
...
+    // with the supplied ID already exists.
+
+    ProjectEntity projectEntity =
+        ProjectEntity.fromRepositoryIdOrNew(firstMockProject.repositoryId);
+
+    assertEquals(firstMockProject, projectEntity);
+  }
+
+  @Test
+  public void fromNewRepositoryId() {
```
Obinnabii, line 152: Are you testing the ```Servlet``` and the ```Entity``` in the same file? maybe these should be separate files?

Richie78321: I'm testing the helper functions included in the `ProjectEntity.java` class. One is the `ProjectEntity.queryFromRequest` seen above and the other is this function, which is used in the mentors servlet. The intention is not to test the servlet here, but just the unit of functionality in `ProjectEntity` that the servlet relies on.

src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+  @Before
+  public void mockResponseSetUp() throws IOException {
+    mockResponse = mock(HttpServletResponse.class);
+    responseStringWriter = new StringWriter();
+    when(mockResponse.getWriter()).thenReturn(new PrintWriter(responseStringWriter));
+  }
+
+  @After
+  public void objectifyTearDown() {
```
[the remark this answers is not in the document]
Richie78321: Sgtm. I renamed this to `objectifyEndSession` and I renamed `objectifySetUp` to `objectifyStartSession` to keep it consistent.

src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  @After
+  public void objectifyEndSession() {
+    session.close();
+    helper.tearDown();
+  }
+
+  @Test
+  public void queryByIds() throws IOException {
```
billyeh, line 85: I think we can give the test more descriptive names, googling "test naming" or "Java test naming" gives lots of good examples.

I'm used to tests having names like `testqueryFromRequestReturnsCorrectProjects` for example. So I can read the test name and understand what it's supposed to do, without reading the comment.

Richie78321: Okay good to know. I'll do a pass over all of my tests with this in mind!

src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
```diff
@@ -0,0 +1,161 @@
...
+    // Expect to receive the ProjectEntities with the corresponding repository IDs.
+
+    HttpServletRequest request = mock(HttpServletRequest.class);
+    when(request.getParameterValues(ProjectEntity.PROJECT_ID_PARAM))
+        .thenReturn(new String[] {firstMockProject.repositoryId, secondMockProject.repositoryId});
+
+    Collection<ProjectEntity> queryResult = ProjectEntity.queryFromRequest(request, mockResponse);
+    Collection<ProjectEntity> expected = Arrays.asList(firstMockProject, secondMockProject);
+    assertNotNull(queryResult);
```
billyeh, line 94:
```suggestion

    assertNotNull(queryResult);
```
A test can generally follow the pattern "setup" "perform action under test" and "check correctness" and I like to structure it that way to make it easy to read. Here it's mostly good, I would just add a newline after performing the "action under test" before checking correctness.

Richie78321: Thanks for the tips. I agree with the suggestion and will apply it!

src/test/java/com/google/opensesame/projects/ProjectEntityTest.java
```diff
@@ -0,0 +1,162 @@
...
+    HttpServletRequest request = mock(HttpServletRequest.class);
+    when(request.getParameterValues(ProjectEntity.PROJECT_ID_PARAM))
+        .thenReturn(new String[] {firstMockProject.repositoryId, secondMockProject.repositoryId});
+
+    Collection<ProjectEntity> queryResult = ProjectEntity.queryFromRequest(request, mockResponse);
+    Collection<ProjectEntity> expected = Arrays.asList(firstMockProject, secondMockProject);
+
+    assertNotNull(queryResult);
+    assertTrue(queryResult.size() == expected.size() && queryResult.containsAll(expected));
+    verify(mockResponse, never()).setStatus(anyInt());
```
billyeh, line 97: Seeing this code here makes me wonder if all our code is in the right place? IMO the *Entity classes are for interacting with Datastore and shouldn't have servlet concepts like Requests and Responses in them. It seems like the responsibilities are a little mixed. 

The fact that we have to check that `setStatus` is called instead of looking at a Response returned is a sign of that, to me. Not a huge deal, something I think might be worth refactoring in the future. The servlet should handle the request parsing and responses and all of that, and just pass the IDs to the Entity class.

Richie78321: I completely agree. I'm thinking we could refactor this by introducing something like the [optional result](https://github.com/Richie78321/google-step/blob/master/portfolio/src/main/java/com/google/sps/helper/ValidationResult.java) that I introduced in my portfolio all those weeks ago. The main reason that I had given the responsibility of error responses to the individual functions was so that context wouldn't be lost from outside of the function if an error occurred, but with some sort of wrapper class we could preserve all of that context! I'll be adding this to the list of issues on the alpha project.

PR #143, opened 2020-07-27
Add tests for the AuthButton

Richie78321:
Add tests for the `AuthButton` in each possible state of the authorization data.

* Extract and export the links used in the `AuthButton` for easier access when testing.
* Add `AuthButton` tests.

Also, thanks @Sami-2000 for getting this testing code started before handing it off to me!


src/main/webapp/js/tests/AuthButton.test.js
```diff
@@ -0,0 +1,84 @@
...
+  USER_SIGNUP_URL,
+} from '../common/react/components/AuthButton.js';
+import '@testing-library/jest-dom/extend-expect';
+import {render, screen} from '@testing-library/react';
+
+/**
+ * Mock auth data representing when a user is authorized and has created a
+ * profile.
+ */
+const mockAuthData1 = {
```
billyeh, line 12: We can give these constants more descriptive names like `AUTHORIZED_WITH_PROFILE_AUTH_DATA`

Richie78321: I agree, that definitely makes it more readable. Will change this now.

PR #144, opened 2020-07-27
Add function for mocking GitHub API interface

Richie78321:
Add a function for mocking the GitHub API interface while testing.

### DIFF
--- a/src/main/java/com/google/opensesame/github/GitHubGetter.java
+++ b/src/main/java/com/google/opensesame/github/GitHubGetter.java
@@ -56,10 +56,10 @@ public final class GitHubGetter {
   }
 
   /**
-   * Sets the GitHub API interface to the provided instance so that all of the following calls
-   * to getGitHub() will return this instance. This is intended to be used for mocking the GitHub
-   * API interface while testing and should not be elsewhere.
-   * 
+   * Sets the GitHub API interface to the provided instance so that all of the following calls to
+   * getGitHub() will return this instance. This is intended to be used for mocking the GitHub API
+   * interface while testing and should not be elsewhere.
+   *
    * @param mockGitHub
    */
   public static void mockGitHub(GitHub mockGitHub) {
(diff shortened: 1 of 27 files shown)


src/main/java/com/google/opensesame/github/GitHubGetter.java
```diff
@@ -54,4 +54,24 @@ public static GitHub buildGitHubInterface(String clientId, String clientSecret)
...
   }
+
+  /**
+   * Sets the GitHub API interface to the provided instance so that all of the following calls to
+   * getGitHub() will return this instance. This is intended to be used for mocking the GitHub API
+   * interface while testing and should not be elsewhere.
+   *
+   * @param mockGitHub
+   */
+  public static void mockGitHub(GitHub mockGitHub) {
```
josephdhurst, line 65: call this `setGitHubForTests` because that is more descriptive of what this function is doing. Otherwise when I look at testing code and see `mockGitHub` I'll be confused since I'd expect a function called `mockGitHub` to actually create the mock.

src/main/java/com/google/opensesame/github/GitHubGetter.java
```diff
@@ -54,4 +54,24 @@ public static GitHub buildGitHubInterface(String clientId, String clientSecret)
...
+  }
+
+  /**
+   * Removes the mocked GitHub API interface so that the following calls to getGitHub() will no
+   * longer return the provided instance. This is intended to be used for mocking the GitHub API
+   * interface while testing and should not be elsewhere.
+   */
+  public static void removeGitHubMock() {
+    gitHub = null;
+  }
```
josephdhurst, line 76: Will we ever need to call this function? I'd assume that we would only set the mock for testing purposes and that we'd be able to use the same mock throughout testing.

Obinnabii:
The Mock GitHub Object was added from PR#145

src/main/java/com/google/opensesame/github/GitHubGetter.java
Richie78321: That's a good point. I had imagined that the best practice with mocking would be to clean up the mock after the tests using the mock had concluded, but we do intend to use the mocks in every test that uses the GitHub getter, so maybe this should be removed?

src/main/java/com/google/opensesame/github/GitHubGetter.java
Richie78321: Solid point here. Thanks, will change.

PR #145, opened 2020-07-27
create GitHub mock object

Obinnabii:
- Mock the ```GitHub API``` with ```Mockito```.
- Mock ```GHMyself``` with ```Mockito```.

look at PR #144 to see how the mock will be used

### DIFF
--- a/src/test/java/com/google/opensesame/mockObjects/MockGHObjects.java
+++ b/src/test/java/com/google/opensesame/mockObjects/MockGHObjects.java
@@ -2,14 +2,18 @@ package com.google.opensesame.mockObjects;
 
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
+import static org.mockito.Matchers.anyString;
 
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.ArrayList;
 import java.util.List;
+
+import org.kohsuke.github.GHMyself;
 import org.kohsuke.github.GHRepository;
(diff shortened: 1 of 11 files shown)


src/test/java/com/google/opensesame/mockObjects/MockGHObjects.java
```diff
@@ -109,6 +115,52 @@ public static GHRepository createMockRepository() throws IOException {
...
+   *     <p>NOTE: all functions return fixtures defined in this class
+   *
+   * @return a mocked instance of a GitHub
+   * @throws IOException
+   */
+  public static GitHub createMockGitHub() throws IOException {
+    GitHub mockGitHub = mock(GitHub.class);
+    when(mockGitHub.getMyself()).thenReturn(createMockMyself());
+    when(mockGitHub.getUser(anyString())).thenReturn(createMockUser());
+    when(mockGitHub.getRepositoryById(anyString())).thenReturn(createMockRepository());
```
Richie78321, line 163: Do you think it might be worth using the ID string provided in `getUser` and `getRepositoryById` as the ID of the mocks? Otherwise I worry about some assumptions possibly being broken where we assume the ID of the user or repo returned from one of these getters is the same as the one provided as a parameter. This can be accomplished using `thenAnswer` instead of `thenReturn` as explained [here](https://stackoverflow.com/a/36627077/9017299).

Obinnabii: I was thinking about that actually, I just made them more specific.

PR #147, opened 2020-07-27
Add Redirect functionality to the Dashboard page

Obinnabii:
- Add a ```script``` to redirect unauthorized users to the ```dashboard``` 
- Add ```fetch``` error handling to the ```dashboard```

### DIFF
--- a/src/main/webapp/dashboard.js
+++ b/src/main/webapp/dashboard.js
@@ -163,7 +163,7 @@ function createRowElement() {
  * Call functions to populate page sections with data.
  */
 async function setUpPage() {
-const user = await getUser();
+  const user = await getUser();
   createAboutMe(user);
 }
 
--- a/src/main/webapp/dashboard.js
+++ b/src/main/webapp/dashboard.js
@@ -1,4 +1,4 @@
-import {getUser} from './user.js';
+import { getUser } from './user.js';
 import {
   standardizeFetchErrors,
   makeRelativeUrlAbsolute,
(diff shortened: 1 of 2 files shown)


src/main/webapp/dashboard.js
```diff
@@ -202,8 +202,12 @@ function getProject(projectID) {
  * Call functions to populate page sections with data.
  */
 async function setUpPage() {
-  const user = await getUser();
-  // TODO: redirect to sign up/in page if the user is not found.
+  const user = await getUser()
+      .catch((errorResponse) => {
+        console.error(
+            `Error ${errorResponse.statusCode}: ${errorResponse.message}`);
+        alert(errorResponse.userMessage);
```
billyeh, line 209: do we want to display an alert? Technically we don't expect this code to be hit, since they should be redirected first right?

Obinnabii: That's a good point! The redirect is a bit slow sometimes and I thought that maybe displaying a message to explain what is going on might be useful for the user. It is kind of pointless though so I'll get rid of it.

Obinnabii: We actually got a couple bad reviews from this feature haha. Good catch!

src/main/webapp/dashboard.js
```diff
@@ -202,8 +202,12 @@ function getProject(projectID) {
...
+      });
```
Obinnabii, line 210:
```suggestion
const user = await getUser();
```

Issue #150, opened 2020-07-28
Extract error responses with wrapper class


Richie78321:
A pattern has arisen where helper functions used by servlets, like those in `ProjectQuery`, will take the servlet response as an argument such that errors can be sent back to the client with all of the necessary context from the helper function. To keep concerns separate, should refactor these helper functions to use a wrapper class on the return value instead.

PR #151, opened 2020-07-28
Add backend support for filters to project queries from servlet.

Richie78321:
Add backend support for filters to project queries coming from servlet requests.

Objectify handles filters [like this](https://www.javadoc.io/static/com.googlecode.objectify/objectify/6.0.6/com/googlecode/objectify/cmd/Query.html#filter(java.lang.String,java.lang.Object)): 
```java
ofy().load().type(ProjectEntity.class).filter("numMentors >", 2);
```
So servlet requests are URL-encoded with filter parameter(s) that take the form of: `field comparator value`. For example: `numMentors >= 2`

Which are then parsed in `QueryFilter.fromString` into `QueryFilter` objects that are then used to conduct the query.

This functionality is being added to enable frontend searching for projects. There will also be a separate PR with the backend support for ordering the project queries.

* Move `ProjectEntity` querying helper functions to a separate class `ProjectQuery` for better separation of concerns.
* Use reflection to determine which `ProjectEntity` fields can be used for filtering based on which fields have the `@Index` annotation; [according to the Objectify wiki](https://github.com/objectify/objectify/wiki/Queries#queries-and-indexes):

> Queries only return values in an index. If an entity field is not indexed, queries on that field will return no results.

* Add `QueryFilter` and `QueryFilter.fromString` to parse the query filters coming from servlet requests.
* Move `ProjectEntityTest.java` to `ProjectQueryTest.java`.
* Add tests for the `QueryFilter` in `QueryFilterTest.java`

_Note: as indicated in issue #150, there will soon be a refactoring to remove the need to take the servlet response as an argument in helper functions like `QueryFilter.fromString`._


Richie78321:
Need to resolve conflict with test and then will reopen

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,94 @@
...
+   * @throws IOException
+   */
+  public static QueryFilter fromString(String QueryFilterString, HttpServletResponse response)
+      throws IOException {
+    if (!QueryFilterString.matches(FILTER_QUERY_REGEX)) {
+      ErrorResponse.sendJsonError(
+          response,
+          "Invalid filter query.",
+          HttpServletResponse.SC_BAD_REQUEST,
+          "Unable to query for projects. Please try again later.");
```
billyeh, line 45: Is "Please try again later" an error message we want to send here? If it doesn't work now, it's unlikely to work in the future, right? Since this isn't a "server is overloaded" kind of message.

Here and below.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,127 @@
...
+   */
+  public static ProjectEntity fromRepositoryIdOrNew(String repositoryId) {
+    ProjectEntity projectEntity = ofy().load().type(ProjectEntity.class).id(repositoryId).now();
+    if (projectEntity == null) {
+      projectEntity =
+          new ProjectEntity(repositoryId, new ArrayList<String>(), new ArrayList<String>());
+    }
+
+    return projectEntity;
+  }
```
billyeh, line 126: Is there a reason to move this function? Seems like it belongs in the ProjectEntity class to me.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,127 @@
...
+  /**
+   * Gets the query filters from a servlet request. Returns null if any errors are encountered while
+   * parsing the query filters. See {@link com.google.opensesame.projects.QueryFilter.fromString}
+   * for more information on how QueryFilters are parsed.
+   *
+   * @param HttpServletRequest The request to get the filter queries from.
+   * @param HttpServletResponse The servlet response to send errors to.
+   * @return Returns the list of QueryFilters or null if an error was encountered while parsing.
+   */
+  private static List<QueryFilter> getQueryFiltersFromRequest(
```
billyeh, line 46: Just an idea, not sure why `getQueryFiltersFromRequest` or `queryFromRequest` has to return `null` in any case. Right now the errors are all handled by the ErrorResponse, so returning null is just another special case the callers will have to handle. Why not just raise the exception? Or even just return an empty list or collection.

src/main/java/com/google/opensesame/projects/QueryFilter.java
Richie78321: Haha that makes sense. Admittedly I wrote that because it sounded professional, but what you said makes sense and I'll remove it.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: This is definitely a symptom of the separation of concerns issue we identified in a previous PR (#150). This is definitely not ideal and could be described as a "hacky" way to indicate to the caller of the method that an error has occurred and a response has been sent.

The reason this needs to be differentiated from returning an empty list is that an empty list indicates to the caller/client that their query returned no results, versus a query that could not be completed being unable to return results at all.

So I'm thinking the best course of action is to get working on issue #150, move errors to a wrapper class, and avoid having to use this `null` return value in any of these methods.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: I think my reasoning was that it is technically "querying" the projects datastore, but on second thought I agree that it makes more sense in the other class. I'll move this class and its tests.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: Actually, on the "Why not just raise the exception," do you think it'd be better to refactor this to use a wrapper class similar to the one used [in my portfolio](https://github.com/Richie78321/google-step/blob/master/portfolio/src/main/java/com/google/sps/helper/ValidationResult.java), or do you think raising a custom exception would be a better and more native solution?

Richie78321: It seems like raising an exception should be saved for [truly exceptional cases](https://stackoverflow.com/a/77361/9017299), and in this case I think it's best to assume that the string can be invalid, so I'm thinking wrapper class, but would like your opinion when you get the chance.

billyeh: I think we're in agreement, this class shouldn't have to know about `HttpServletRequest` or `HttpServletResponse` objects.

Error wrapper or exception is fine. I was leaning toward throwing an exception. I don't see "truly exceptional cases" in the answer you linked? I do get that reading from the second-most upvoted answer.

The first answer (linked) talks about throwing exceptions when pre-assumptions are broken. Seems like that's exactly what's happening here, if I have a function to parse a query from a string then it seems reasonable to me to expect that string to be in a particular format. And if it isn't, that would be where an exception comes in.

A plus of exceptions is that you get the stack trace for free.

Richie78321: Okay, sounds good. I'll refactor this.

Richie78321: @billyeh Just finished refactoring. I love the result! Much cleaner in usage and in testing. Thanks so much for the suggestion.

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+          "Cannot parse the comparison value '"
+              + comparisonValue
+              + "' for the field '"
+              + filterFieldName
+              + "'.",
+          "Unable to query for projects.",
+          HttpServletResponse.SC_BAD_REQUEST);
+    }
+
+    String comparator = splitFilterRequest[1];
```
josephdhurst, line 77: nit: if you decide not to follow my comment about moving `comparisonValue` and `comparator` right after `splitFilterRequest`, consider moving this variable definition before `comparisonValue` so that the indices of splitFilterRequest are in order. As it stands the order is 0, 2, 1 which is slightly confusing.

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+    return new QueryFilter(filterFieldName + " " + comparator, comparisonObject);
```
josephdhurst, line 78: If possible create separate functions for validation that throw `ServletValidationException`s if they fail (in my example I've named them `validateFilterFieldName` and `createComparisonObject`). This will break up `QueryFilter.fromString` into smaller and more manageable functions, making the code more modular and easier to read.
```suggestion
    // should throw a ServletValidationException if it fails
    validateFilterFieldName(filterFieldName);
    
    String comparator = splitFilterRequest[1];

    String comparisonValue = splitFilterRequest[2];
    // should throw a ServletValidationException if it fails
    Object comparisonObject = createComparisonObject(comparisonValue);

    return new QueryFilter(filterFieldName + " " + comparator, comparisonObject);
```

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+            .filter((field) -> field.getName().equals(filterFieldName))
+            .findFirst();
+    if (!filterField.isPresent()) {
+      throw new ServletValidationException(
+          "Cannot filter by the field '" + filterFieldName + "'.",
+          "Unable to query for projects.",
+          HttpServletResponse.SC_BAD_REQUEST);
+    }
+
+    String comparisonValue = splitFilterRequest[2];
```
josephdhurst, line 60: using `comparison` in this variable name confuses me a lot. It took me a couple minutes to realize that this variable is the comparator. Just call this variable `value` to stay consistent with your docstring that states that `QueryFilterString` will come in the format "fieldName \<comparator\> value".
```suggestion
    String value = splitFilterRequest[2];
```

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,118 @@
...
+
+    return queryFilters;
+  }
+
+  /**
+   * Queries for ProjectEntities in the datastore based on a servlet request.
+   *
+   * @param request The servlet request defining the query.
+   * @param response The servlet response to send errors to.
+   * @return Returns a collection of ProjectEntities from the query or null if the query was
```
josephdhurst, line 68: does this function return null or does it just raise an error?

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,118 @@
...
+          }
+        }
+
+        projectEntities = intermediateIntersection;
+      } else {
+        projectEntities = ofy().load().type(ProjectEntity.class).list();
+      }
+    }
+
+    return projectEntities;
```
josephdhurst, line 116: It seems like there is a lot going on in this function, so I think it should be broken up into smaller functions:
- a function that gets `ProjectEntity`s from projectIds (in my example I call it `getProjectEntitiesByIds`)
- a function that gets `ProjectEntity`s from a queryFilters (in my example I call it `getProjectEntitiesByQueryFilters`)
```suggestion
    String[] projectIds = request.getParameterValues(ProjectEntity.PROJECT_ID_PARAM);
    if (projectIds != null) {
      return getProjectEntitiesByIds(projectIds);
    } else {
      List<QueryFilter> queryFilters = getQueryFiltersFromRequest(request);
      return getProjectEntitiesByQueryFilters(queryFilters);
    }
```

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,118 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.List;
+import java.util.stream.Collectors;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+/** Helper functions for querying the projects Datastore. */
+public class ProjectQuery {
+  public static final String FILTER_QUERY_PARAM = "filter";
```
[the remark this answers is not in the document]
Richie78321: > However, in this case I don't think that this value will ever change since it is part of a well-established API.

Just to clarify, this is being used by my function `ProjectQuery.getQueryFiltersFromRequest`. Would you mind elaborating on what you mean by "part of a well-established API"?

Also, my reasoning for extracting it into a constant was so that future developers wouldn't have to look deep into the code to determine what the URL parameter is for sending filters. Is there maybe another way that I can make this value easy for developers to find?

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+/**
+ * A Datastore filter stored in the form used by Objectify when loading entities from the Datastore.
+ *
+ * <p>See here for how Objectify conducts filtering on a query:
+ * https://www.javadoc.io/static/com.googlecode.objectify/objectify/6.0.6/com/googlecode/objectify/cmd/Query.html#filter(java.lang.String,java.lang.Object)
+ */
+class QueryFilter {
+  public static final String FILTER_QUERY_REGEX = "^[A-Za-z]+ (>|>=|!=|=|<|<=) .+$";
```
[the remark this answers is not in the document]
Richie78321: These are solid points, thanks! I'll update this and keep in mind in the future that scope = additional context.

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+   * <pre>
+   * "numMentors >= 2"
+   * "numInterestedUsers = 0"
+   * </pre>
+   *
+   * </blockquote>
+   *
+   * @param QueryFilterString The filter query string.
+   * @param response The servlet response to send errors to.
+   * @return Returns the QueryFilter or null if there was an error parsing the string.
```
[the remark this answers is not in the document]
Richie78321: Yeah good catch; it seems I forgot to update the documentation after switching to throwing exceptions instead of returning null. I'll do a pass over all of the JavaDoc and comments.

src/main/java/com/google/opensesame/projects/QueryFilter.java
```diff
@@ -0,0 +1,88 @@
...
+      throws IOException, ServletValidationException {
+    if (!QueryFilterString.matches(FILTER_QUERY_REGEX)) {
+      throw new ServletValidationException(
+          "Invalid filter query.",
+          "Unable to query for projects.",
+          HttpServletResponse.SC_BAD_REQUEST);
+    }
+
+    String[] splitFilterRequest = QueryFilterString.split(" ");
+    String filterFieldName = splitFilterRequest[0];
```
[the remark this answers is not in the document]
Richie78321: Okay, that makes sense. I'll move the variable assignments up here. And as for checking that `splitFilterRequest.length` equals 3, that should be guaranteed if the string matches the regex string. Do you think it's worth adding the check anyways?

src/main/java/com/google/opensesame/projects/QueryFilter.java
Richie78321: I agree. Sorry it took so long to read! Will change.

src/main/java/com/google/opensesame/projects/QueryFilter.java
Richie78321: Good point. Will update.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: Apologies, this is due to me forgetting to update the JavaDoc after moving to using exceptions. This will be updated!

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: I agree with you. I just implemented this change and will push shortly. Thanks for all of the help!

src/main/java/com/google/opensesame/projects/QueryFilter.java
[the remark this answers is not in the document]
josephdhurst: Oops, I barely glanced at the regex. Definitely don't add a redundant check.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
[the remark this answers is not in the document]
josephdhurst: I think that it would be fairly easy to just search the file for "getParameterValues" using ctrl + f.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
```diff
@@ -0,0 +1,143 @@
...
+
+      if (intermediateIntersection != null) {
+        intermediateIntersection =
+            filteredProjects.stream()
+                .filter(intermediateIntersection::contains)
+                .collect(Collectors.toList());
+      } else {
+        intermediateIntersection = filteredProjects;
+      }
+    }
```
josephdhurst, line 115: There should be a way to apply multiple filters at once. I'm not super familiar with what is going on but would something like this work?
```suggestion
    List<ProjectEntity> filteredProjects =
          ofy()
              .load()
              .type(ProjectEntity.class)
    for (QueryFilter queryFilter : queryFilters) {
              filteredProjects = filteredProjects.filter(queryFilter.condition, queryFilter.comparisonObject)
    }
    return filteredProjects.list()

File: src/main/java/com/google/opensesame/projects/QueryFilter.java [hunk 14]
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Converts the filter value string to an object of the filter field type.
+   *
+   * <p>The filter field type must have a constructor that takes a single string as an argument for
+   * this conversion.
+   *
+   * @param valueString A string representing the value of the object.
+   * @param filterField The field that this filter is filtering by.
+   * @return Returns the object created from the value string.
```
josephdhurst, line 88: The phrase "the object" is pretty vague. If possible replace "the object" with a more specific description.

src/main/java/com/google/opensesame/projects/ProjectQuery.java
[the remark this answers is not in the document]
Richie78321: Yeah that makes sense. I'll put it into the method, thanks!

src/main/java/com/google/opensesame/projects/QueryFilter.java
Richie78321: Yeah I agree. I'll see about replacing the name for that variable and these comments in the whole document!

src/main/java/com/google/opensesame/projects/ProjectQuery.java
Richie78321: I wish! This is very similar to the code that I had originally written for this functionality, but due to how Datastore works internally you can only make one filter per query to the Datastore. That's why the best option with Datastore is to make multiple queries and determine the intersection of all of the filters. It's not amazingly efficient, but if this ever becomes a problem, we can make use of [projection queries](https://cloud.google.com/appengine/docs/standard/java/datastore/projectionqueries) to pull limited data from Datastore on the intermediate queries.

PR #152, opened 2020-07-28
Fix null name bug

Obinnabii:
This will make a users name, show up as the user's ```GitHub Login``` if the user doesn't have a name set on ```GitHub```

### DIFF
--- a/src/main/java/com/google/opensesame/user/UserData.java
+++ b/src/main/java/com/google/opensesame/user/UserData.java
@@ -37,6 +37,11 @@ public class UserData {
    * the AuthServlet. This is to avoid unauthorised users from contacting our mentors. We want to
    * avoid email spam as much as possible and we do not see the need to share user emails
    * unathorized personnel.
+   * 
+   * <p>Note: If a user's name is not defined on GitHub, this datastructure will
+   * set the user's GitHub login ID as the user's name.
+   * i.e {@code if (name == null);}
+   * {@code name = gitHubID}
    *
    * @param userEntity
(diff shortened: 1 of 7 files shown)


src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -60,6 +63,9 @@ public UserData(UserEntity userEntity) throws IOException {
     image = userGitAccount.getAvatarUrl();
     location = userGitAccount.getLocation();
     name = userGitAccount.getName();
+    if (name == null) {
+      name = gitHubID;
+    }
```
Richie78321, line 68: Could we instead use `name = userGitAccount.getLogin()` here? We know that a user will have a GitHub username, and that seems like a much more human-readable backup than using the numerical ID.

[Here](https://github-api.kohsuke.org/apidocs/org/kohsuke/github/GHPerson.html#getLogin()) is the Java GitHub API reference.

Richie78321: Okay just realized that we are storing a user's login as their `gitHubId`. Could we document this or change the naming to be more consistent with GitHub's nomenclature?

Obinnabii: Yeah, the Login is the only identifier we get from the API. Sadly, the nomenclature has been widely adopted across our website. To improve consistency with the GitHub API, I'll try to change all the names. This might have adverse effects on the already rolled out website though. We might have to clear Datastore.

PR #156, opened 2020-07-30
Add Test for Dashboard

Obinnabii:
- Test that User information is displayed on the dashboard when it is provided by the ```UserServlet```. 
- remove email checks on the dashboard because they have become redundant with the recent authorization checks.
- Switch from using ```.innerText``` to ```.innerHTML```. This makes the code more testable. ```Jest-dom``` is only really able to detect ```innerHTML``` text. Since they essentially work the same way, it wasn't too big of a deal. It is something to keep in mind when writing ```JavaScript DOM-manipulating``` code. An alternative to ```innerHTML``` is to use ```textNodes```. ```textNodes``` help prevent XSS attacks from user inputted text. 
- Pass in the ```document``` object as a parameter to ```createAboutMe``` (another decision to improve testability). Honestly, this raises concerns for me about how to truly separate concerns. It feels like the code would be easier to test and more manageable if all the `HTML Elements` that get altered by `JS DOM` functions were `JS` generated. With the current model, tests need to instantiate an `HTML Document` object that loosely resembles whatever file the script will be run on.  

This whole Jest thing was a huge learning experience for me. I iterated over this code a couple times.


src/main/webapp/dashboard.js
```diff
@@ -27,44 +27,59 @@ import {getUser} from './user.js';
...
   userNameAndLocationElement.innerHTML = user.name + '<br>';
   if (user.location) {
     const userLocation = createLocation(user.location);
     userNameAndLocationElement.append(userLocation);
   }
 
+  const userGithubButton = document.getElementById(USER_GITHUB_ID);
   userGithubButton.href = user.gitHubURL;
+  const userEmailButton = document.getElementById(USER_EMAIL_ID);
   if (!user.email) {
```
Sami-2000, line 74: Nit Q: Could a user ever have access to the dashboard page and not have an email?

src/main/webapp/dashboard.js
```diff
@@ -133,7 +156,8 @@ function addProjectTag(project, projectDiv) {
  */
 function createLocation(location) {
```
Sami-2000, line 157: Q: Would functions like this, which are called by createAboutMe and use the 'document' object, require 'document' to be passed in as a param?

src/main/webapp/dashboard.test.js
```diff
@@ -0,0 +1,132 @@
...
+  image: 'http://localhost/image.jpg',
+  gitHubID: 'mockGitID',
+  gitHubURL: 'http://localhost/mockGitID',
+};
+
+// Note: Update this when the dashboard gets changed
+/**
+ * Function to replicate the DOM of Dashboard.html
+ */
+function resetDocumentBody() {
```
Sami-2000, line 38: Q: Is there a way for you to get this block from your html file?

src/main/webapp/dashboard.test.js
```diff
@@ -0,0 +1,132 @@
...
+  });
+});
+
+describe('About me card of a user with missing optional fields', () => {
+  beforeAll(() => {
+    resetDocumentBody();
+    dashboard.createAboutMe(mockUserMissingOptionalFields, document);
+  });
+
+  it('hides the email button', () => {
```
Sami-2000, line 124: Q: Once again, I am just wondering if it is ever possible to successfully enter the dashboard page and retrieve user info without that user having input an email.

src/main/webapp/dashboard.js
Obinnabii: No, I don't think so. I added this in case we would make this into a reusable breakout page. But we recently went in a different direction so I'll remove the email and GitHubURL stuff

src/main/webapp/dashboard.js
Obinnabii: for calls to ```createElement,``` it should be fine. It only really matters if they are adding to the ```DOM``` i think

src/main/webapp/dashboard.test.js
Obinnabii: I looked into it and I couldn't figure out a good way. [this](https://dev.to/ms314006/use-jest-write-unit-testing-for-dom-manipulation-3n6c) was the resource I ended up using to decide how to write my tests. I'm very open to changing how the tests work, this feels very sweaty to me as it is.

src/main/webapp/dashboard.test.js
```diff
@@ -0,0 +1,132 @@
...
+// Note: Update this when the dashboard gets changed
+/**
+ * Function to replicate the DOM of Dashboard.html
+ */
+function resetDocumentBody() {
+  document.body.innerHTML = `
+  <div id="about-me" class="container tab-pane active"><br>
+    <div class="card card-holder col-12 ">
+      <div id="about-me-card-body" class="card-body">
+        <img id="user-image" class="card-img-top img-thumbnail p-2 m-2
```
billyeh, line 43:
```suggestion
        <img id="${dashboard.USER_IMAGE_ID}" class="card-img-top img-thumbnail p-2 m-2 
```
Javascript template strings have interpolation, might as well use the exported IDs here as well.

src/main/webapp/dashboard.test.js
```diff
@@ -0,0 +1,128 @@
...
+  document.body.innerHTML = `
+  <div id="about-me" class="container tab-pane active"><br>
+    <div class="card card-holder col-12 ">
+      <div id="about-me-card-body" class="card-body">
+        <img id="user-image" class="card-img-top img-thumbnail p-2 m-2 
+            mx-auto d-block user-img" src="#" alt="Card image">
+        <h4 id="user-name-location" class="card-title dark-emph"></h4>
+        <p id="user-bio" class="card-text"></p>
+        <a id="user-github" href="#" class="btn btn-primary btn-emphasis" 
+            role="button">
```
billyeh, line 49:
```suggestion
        <a id="user-github" href="#">
```
Also, I'd consider getting rid of the classes/fields we don't care about for testing purposes. It'll make things simpler.

PR #157, opened 2020-07-30
Render project README on project breakout page

Richie78321:
Add support for rendering the project README onto the project breakout page.

* Add backend support for sending the README.md raw and HTML URLs to be used on the frontend.
* Add frontend support for fetching the raw markdown, parsing it into HTML using [Marked](https://github.com/markedjs/marked), sanitizing it to defend against XSS using [DOMPurify](https://github.com/cure53/DOMPurify), and rendering it to the project breakout React component.
* Add a local copy of the [GitHub markdown CSS](https://github.com/sindresorhus/github-markdown-css).

### DIFF
--- a/src/main/java/com/google/opensesame/projects/ProjectData.java
+++ b/src/main/java/com/google/opensesame/projects/ProjectData.java
@@ -167,8 +167,8 @@ public class ProjectData {
   private transient boolean repositoryReadmeLoaded = false;
   /**
    * Gets the repository README.md file or returns a cached version if it exists. If the repository
-   * does not have a README.md file, returns null. 
-   * 
+   * does not have a README.md file, returns null.
+   *
    * @return Returns the repository README.md file or returns null if the repository does not have
    *     one.
    */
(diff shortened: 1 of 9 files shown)


src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -26,6 +28,8 @@
   private Integer numMentors = null;
   private String repositoryId = null;
   private List<UserData> mentors = null;
+  private String readmeRawUrl = null;
+  private String readmeHtmlUrl = null;
```
Obinnabii, line 32: nit: these names are kind of confusing. It sounds a lot like Raw is describing the URL and not the readme. These seems like a difficult set of fields to name so it's not a big deal. Maybe we could stick to the API's convention of ```Downloadable vs HTML```

src/main/java/com/google/opensesame/projects/ProjectData.java
```diff
@@ -158,4 +162,61 @@ public String getRepositoryId() {
 
     return Collections.unmodifiableList(mentors);
   }
+
+  private transient GHContent repositoryReadme = null;
+  private transient boolean repositoryReadmeLoaded = false;
```
josephdhurst, line 167: is there a reason why `repositoryReadmeLoaded` is transient?

Richie78321: Yes. It's not a piece of data that I feel is relevant to the client and is mostly there to avoid having to reload the repository README from the GitHub API more than once.

PR #158, opened 2020-07-30
Sort mentors.

Sami-2000:
When the user is logged in and has a profile, sort the mentors on the mentor search page by # of shared interests.

- Add a 'compatibility' function to the UserData class. Currently this function just returns the number of shared interest tags between two users.
- Add to the doGet function of the MentorsServlet. Now, before returning the list of mentors, first check if the user is authenticated and has a profile in datastore. If so, sort the mentor list based on the UserData 'compatibility' function.

### DIFF
--- a/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
+++ b/src/main/java/com/google/opensesame/servlets/MentorsServlet.java
@@ -37,24 +37,26 @@ public class MentorsServlet extends HttpServlet {
       mentors.add(new UserData(mentorEntity));
     }
 
-    //If the user is logged in and has a profile, sort mentors by common interest.
+    // If the user is logged in and has a profile, sort mentors by common interest.
     User loggedInUser = AuthServlet.getAuthorizedUser();
     if (loggedInUser != null) {
       String userID = loggedInUser.getUserId();
       UserEntity curUser = ofy().load().type(UserEntity.class).id(userID).now();
       if (curUser != null) {
         final UserData curUserData = new UserData(curUser);
-        Collections.sort(mentors, new Comparator<UserData>() {
-          @Override
-          public int compare(UserData u1, UserData u2) {
-            Integer u1Compatibility = new Integer(u1.compatibility(curUserData));
-            Integer u2Compatibility = new Integer(u2.compatibility(curUserData));
-            return u2Compatibility.compareTo(u1Compatibility);
-          }
-        });
+        Collections.sort(
+            mentors,
+            new Comparator<UserData>() {
+              @Override
+              public int compare(UserData u1, UserData u2) {
+                Integer u1Compatibility = new Integer(u1.compatibility(curUserData));
+                Integer u2Compatibility = new Integer(u2.compatibility(curUserData));
+                return u2Compatibility.compareTo(u1Compatibility);
+              }
+            });
       }
     }
-    
+
     String jsonMentors = new Gson().toJson(mentors);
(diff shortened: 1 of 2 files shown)


src/main/java/com/google/opensesame/servlets/MentorsServlet.java
```diff
@@ -26,14 +29,34 @@
...
+      UserEntity curUser = ofy().load().type(UserEntity.class).id(userID).now();
+      if (curUser != null) {
+        final UserData curUserData = new UserData(curUser);
+        Collections.sort(
+            mentors,
+            new Comparator<UserData>() {
+              @Override
+              public int compare(UserData u1, UserData u2) {
+                Integer u1Compatibility = new Integer(u1.compatibility(curUserData));
+                Integer u2Compatibility = new Integer(u2.compatibility(curUserData));
```
Richie78321, line 53: This _might_ fall into premature optimization, but it might be worth pre-computing the compatibility scores of each of the mentors before hand and storing them in a HashMap so it doesn't need to be recomputed for every comparison.

src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -139,4 +139,17 @@ private void loadProjects(ArrayList<String> projectIds) throws IOException {
...
+
+  /**
+   * @param {UserData} user2 the user to compare to
+   * @return an integer 'compatibility score' which represents how many shared interests two users
+   *     have.
+   */
+  public int compatibility(UserData user2) {
+    ArrayList<String> user1Interests = (ArrayList) this.interestTags.clone();
+    ArrayList<String> user2Interests = (ArrayList) user2.interestTags.clone();
+    user1Interests.retainAll(user2Interests);
```
Richie78321, line 151: nit: [Java streams can be used here](https://www.baeldung.com/java-lists-intersection#strings) to avoid having to clone the lists before calculating the intersection. Not that big of a deal, but I thought it was worth a mention as streams are quite clean!

src/main/java/com/google/opensesame/user/UserData.java
```diff
@@ -139,4 +139,17 @@ private void loadProjects(ArrayList<String> projectIds) throws IOException {
...
+    System.out.println(user1Interests.size());
```
Richie78321, line 152: Is this intended to be here? If so, it might be worth adding a TODO to remove this for production.

src/main/java/com/google/opensesame/servlets/MentorsServlet.java
josephdhurst: I think that is a premature optimization and we should revisit it later if necessary, but not include it in this PR.

src/main/java/com/google/opensesame/user/UserData.java
josephdhurst: If we're going to keep this line in print some context along with the int so that we know what it is. If we just have print statements that print numbers or strings without additional context it makes the logs unreadable when 10+ print statements from different functions get called.

Sami-2000: So sorry, I definitely meant to delete this! Joe, thank you for the advice for future print statements.

src/main/java/com/google/opensesame/user/UserData.java
Sami-2000: Those look great for this! Thanks for sharing.

PR #159, opened 2020-07-31
Add deployment steps to the readme & firebase SDK import change

Richie78321:
Add deployment steps to the readme, and replace the full Firebase SDK with the individual SDKs to avoid having to do so during deployment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ A user is only considered to have an account on Open Sesame if they have complet
 
 ## Deployment Steps
 When deploying Open Sesame to GCP, please follow these steps:
-1. Replace all imported JS scripts with their production counterparts. The individual instructions for each script will be located above the script tags in the HTML files where they are imported. This includes but is not limited to the React scripts and Firebase scripts.
+1. Replace all imported JS scripts with their production counterparts. The individual instructions for each script will be located above the script tags in the HTML files where they are imported. This includes but is not limited to the React and PropType scripts.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -117,6 +117,14 @@ User signup consists of three parts:
 
 A user is only considered to have an account on Open Sesame if they have completed all three steps outlined above.
 
+## Deployment Steps
+When deploying Open Sesame to GCP, please follow these steps:
+1. Replace all imported JS scripts with their production counterparts. The individual instructions for each script will be located above the script tags in the HTML files where they are imported. This includes but is not limited to the React and PropType scripts.
+2. Look over the codebase TODOs and disable code that is intended to be disabled for production.
+3. Add the associated GitHub organization's `client-id` and `client-secret` to the App Engine environment variables located in `open-sesame/src/main/webapp/WEB-INF` to [increase the GitHub rate limit of the application](https://developer.github.com/v3/#oauth2-keysecret).
+4. Add the GCP project ID to the `pom.xml` file.
+5. Deploy!
```
josephdhurst, line 126: This process seems incredibly complicated and error-prone. We should talk about deployment automation. Even though we probably don't have time to do it for this project, it would be interesting to look into because it could save you a lot of time in future projects.

Richie78321: Sounds good to me! I'll add it to the list of talking points for today's sync.

PR #160, opened 2020-07-31
Use stretched link.

Sami-2000:
Given that all users will now have a name field, link from mentor card to mentor breakout page using a stretched link of the name div.
This prevents all text on the mentor card from acting like a link (being blue, underlining, etc.).

### DIFF
--- a/src/main/webapp/mentors.js
+++ b/src/main/webapp/mentors.js
@@ -67,11 +67,7 @@ function getMentors() {
  * @return {HTMLElement} mentorContainer
  */
 export function createMentorElement(mentor) {
-  const mentorContainer = document.createElement('a');
-  const breakoutUrl = new URL('/mentor_breakout.html', window.location.origin);
-  console.log(mentor.userID);
-  breakoutUrl.searchParams.append('mentorID', mentor.userID);
-  mentorContainer.href = breakoutUrl;
+  const mentorContainer = document.createElement('div');
   mentorContainer.className = 'p-1 col-lg-4';
 
   const mentorCard = document.createElement('div');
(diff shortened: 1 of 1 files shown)


src/main/webapp/mentors.js
```diff
@@ -83,9 +79,19 @@ export function createMentorElement(mentor) {
...
+  mentorNameContainer.className = 'stretched-link';
+  const breakoutUrl = new URL('/mentor_breakout.html', window.location.origin);
+  console.log(mentor.userID);
+  breakoutUrl.searchParams.append('mentorID', mentor.userID);
+  mentorNameContainer.href = breakoutUrl;
+  const mentorName = document.createElement('h5');
+  mentorName.className = 'card-title text-primary';
+  mentorName.innerHTML = mentor.name;
+  mentorNameContainer.appendChild(mentorName);
+  mentorCardBody.appendChild(mentorNameContainer);
```
Richie78321, line 93: Q: Is there a reason this was moved into the main function?

Sami-2000: Yes! Because Obi changed the backend so that all users now have a name, I figured it was best dealt with in the 'main' function.

However, if you believe this makes the code less readable, I could very easily create the mentorNameContainer in this function, and keep the creation of the mentorName element seperated into a different function.

Richie78321: Okay! Thanks for the answer. I don't have any real strong opinion on this so feel free to keep it where you think it's most logical.

josephdhurst: I think that this new logic should be broken out into a separate function because:
- It doesn't depend much on the other logic in createMentorElement
- createMentorElement is already a large function and we want to avoid making it larger if possible.

PR #163, opened 2020-08-03
Make GitHub throw an exception when it runs out of API calls

Obinnabii:
- Create a ```RateLimitExceededException``` to signify when the `GitHub Java API` has run out of calls
- Create a `RateLimitHandler` to throw the above exception when the `GitHub Java API` has run out of calls and integrate it into the `Github API`

### DIFF
--- a/src/main/java/com/google/opensesame/github/GitHubGetter.java
+++ b/src/main/java/com/google/opensesame/github/GitHubGetter.java
@@ -1,10 +1,8 @@
 package com.google.opensesame.github;
 
+import com.google.opensesame.util.RateLimitExceededException;
 import java.io.IOException;
 import java.net.HttpURLConnection;
-
-import com.google.opensesame.util.RateLimitExceededException;
-
 import org.kohsuke.github.GitHub;
 import org.kohsuke.github.GitHubBuilder;
 import org.kohsuke.github.RateLimitHandler;
@@ -13,21 +11,17 @@ public final class GitHubGetter {
   private static GitHub gitHub = null;
 
   /** See {@link GitHubRateLimitHandler} */
-  private static GitHubRateLimitHandler rateLimitHandler = 
-      new GitHubRateLimitHandler();
+  private static GitHubRateLimitHandler rateLimitHandler = new GitHubRateLimitHandler();
 
   /**
-   * Gets the GitHub API interface or builds one if there is not one available. If
-   * credentials are available in the system environment variables, the builder
-   * will connect to the GitHub API using those. Otherwise, it will connect
-   * anonymously.
+   * Gets the GitHub API interface or builds one if there is not one available. If credentials are
+   * available in the system environment variables, the builder will connect to the GitHub API using
+   * those. Otherwise, it will connect anonymously.
    *
-   * <p>
-   * In AppEngine, the credentials can be set in appengine-web.xml.
+   * <p>In AppEngine, the credentials can be set in appengine-web.xml.
    *
    * @return Returns the GitHub API interface.
-   * @throws IOException Throws IOException if there is an error building the
-   *                     GitHub interface.
+   * @throws IOException Throws IOException if there is an error building the GitHub interface.
    */
   public static GitHub getGitHub() throws IOException {
     if (gitHub == null) {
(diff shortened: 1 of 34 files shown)


src/main/java/com/google/opensesame/util/RateLimitExceededException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * <p>NOTE: When this Exception is caught, the appropriate error code to throw would be an
+ * HTML_FORBIDDEN (403). This is to stay consistent with the GitHub Restful API convention found <a
+ * href="https://developer.github.com/v3/#rate-limiting">here</a>
+ */
+public class RateLimitExceededException extends IOException {
+
+  /** The default error message to be sent when OpenSesame runs out of API calls. */
+  private static String defaultErrorMessage =
+      "OpenSesame has Exceeded it's " + "available GitHub API calls";
```
Richie78321, line 18: Typos and fixing after auto-linter:
```suggestion
      "Open Sesame has exceeded its available GitHub API calls.";
```

src/main/java/com/google/opensesame/util/RateLimitExceededException.java
```diff
@@ -0,0 +1,23 @@
+package com.google.opensesame.util;
+
+import java.io.IOException;
+
+/**
+ * The exception to be thrown when the GitHub API has run out of calls. This will help us avoid
+ * lengthy stalls when we lose access to GitHub Data i.e. Our API calls have exceeded our current
+ * GitHub hourly call rate limit.
```
Richie78321, line 8: nit: I don't think this information is 100% relevant here. Consider removing this comment, as this information is already present in a more relevant spot on the `GitHubRateLimitHandler`.
```suggestion
 * The exception to be thrown when the GitHub API has run out of calls.
```

src/main/java/com/google/opensesame/util/RateLimitExceededException.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * <p>NOTE: When this Exception is caught, the appropriate error code to throw would be an
+ * HTML_FORBIDDEN (403). This is to stay consistent with the GitHub Restful API convention found <a
+ * href="https://developer.github.com/v3/#rate-limiting">here</a>
```
Richie78321, line 12: Nice!

src/main/java/com/google/opensesame/util/RateLimitExceededException.java
Richie78321, line 18: Consider moving this into the no-arg constructor below, as it doesn't provide much extra functionality or information by being in a constant like this. Take a look at these comments I got from Joe, which were quite helpful for me:

https://github.com/googleinterns/open-sesame/pull/151#discussion_r463924612
https://github.com/googleinterns/open-sesame/pull/151#discussion_r463926351

src/main/java/com/google/opensesame/github/GitHubGetter.java
```diff
@@ -74,4 +86,17 @@ public static void setGitHubForTests(GitHub mockGitHub) {
   public static void removeGitHubMock() {
     gitHub = null;
   }
+
+  /**
+   * Causes the GitHub API to fail with a RateLimitExceededException when it has exceeded the API
+   * calls alotted to it by GitHub. Without this handler, the API would stall until its available
+   * calls are replenished.
```
Richie78321, line 93: nit: This wording is a bit confusing imo. Consider:
```suggestion
   * Causes requests made to the GitHub API to fail with a RateLimitExceededException when the API
   * rate limit has been exceeded. Without this handler, requests to the API would hang until the limit
   * has been refreshed.
```

src/main/java/com/google/opensesame/github/GitHubGetter.java
```diff
@@ -1,12 +1,18 @@
...
+import java.net.HttpURLConnection;
 import org.kohsuke.github.GitHub;
 import org.kohsuke.github.GitHubBuilder;
+import org.kohsuke.github.RateLimitHandler;
 
 public final class GitHubGetter {
   private static GitHub gitHub = null;
 
+  /** See {@link GitHubRateLimitHandler} */
+  private static GitHubRateLimitHandler rateLimitHandler = new GitHubRateLimitHandler();
```
Richie78321, line 14: nit: Knowing that `buildGitHubInterface` is only run once and the `rateLimitHandler` is only used in that function, it might be best to move this to a local variable in `buildGitHubInterface`.

josephdhurst: +1

josephdhurst: It's usually best to keep the scope of variables as limited as possible. This helps prevent them from being unintentionally accessed and modified.

src/main/java/com/google/opensesame/util/RateLimitExceededException.java
```diff
@@ -0,0 +1,23 @@
+package com.google.opensesame.util;
+
+import java.io.IOException;
+
+/**
+ * The exception to be thrown when the GitHub API has run out of calls. This will help us avoid
+ * lengthy stalls when we lose access to GitHub Data i.e. Our API calls have exceeded our current
+ * GitHub hourly call rate limit.
+ *
+ * <p>NOTE: When this Exception is caught, the appropriate error code to throw would be an
```
josephdhurst, line 10: nit:
```suggestion
 * <p>NOTE: When this Exception is caught, the appropriate error code to throw is an
```

src/main/java/com/google/opensesame/util/RateLimitExceededException.java
Obinnabii: Solid point!

Obinnabii: It turns out Java constructors don't allow you to make any declarations before calling the super class constructor. Tbh, the text has minimal utility. should I remove it?

